PR #253, opened 2015-03-13
Make complianceitem plugin extract item data from event message

jvehent:


jvehent:
@ameihm0912: r?

ameihm0912:
Looks good, will also make required changes in mozdef_client for this

PR #261, opened 2015-03-26
Nicer colors and formatting of bug for readability

gdestuynder:


bot/modules/zilla.py
```diff
@@ -71,7 +71,7 @@ def bugzilla_search(self):
             except:
                 return
             for bug in res['bugs']:
-                self.controller.client.msg(self.channel, "[zilla] NEW bug: {summary} {url}{bugid}".format(summary=bug['summary'],
+                self.controller.client.msg(self.channel, "\x038\x02WARNING\x03\x02 \x032\x02NEW\x03\x02 bug: {url}{bugid} {summary}".format(summary=bug['summary'],
```
jeffbryner: no way to get to the awesome colorify function? Should we refactor it?

gdestuynder: we have to either modify upstream IRC code to integrate it in the irc module class, or to make it a separate class/lib that the module can import.
any preference?

jeffbryner: Maybe this? https://github.com/dslackw/colored

gdestuynder:
hmm i would think that the best is to pass any string to the colorifier maybe
so that its consistent

PR #282, opened 2015-06-17
Fix validation checks for RRA plugin (slightly more verbose, too), update submodules

gdestuynder:


gdestuynder:
tested on mozdefqa1

Issue #292, opened 2015-09-29
WebUI: Attackers: Allow search by IP


jeffbryner:
In addition to the filtering by country, date, category add a feature to the attackers screen to allow searching by IP. 

This will help in categorizing/re-categorizing attackers in case of false positives.

yashmehrotra:
Fixed in https://github.com/jeffbryner/MozDef/pull/299. @jeffbryner Kindly close it.

PR #299, opened 2015-10-11
Attackers: Added filter to allow search by IP, fix for #292

526avijitgupta:
Some UI changes have been made to accomodate the IP filter in the `sidenav` Here is the screenshot of the same: 
screenshot from 2015-10-11 19 30 31


yashmehrotra:
I also took the liberty of adding a `Reset Filters` button since it was getting a bit difficult to manage 4 filters.

This is how it looks:
screenshot from 2015-10-13 05 50 40

PR #307, opened 2015-10-21
Fix submodule reference

gdestuynder:
This uses an actually existing submodule reference. I suspect the older reference was made from a local checkout that no longer exists, albeit I can't be sure.
That should fix git clone --recursive of the mozdef repo.

See also http://stackoverflow.com/questions/2155887/git-submodule-head-reference-is-not-a-tree-error if you need to do that in the future.

The tldr is that you checkout the submodule manually to a known-good commit from the upstream project, then commit that. It will create a new submodule reference commit as in this PR. Then you should be good!


gdestuynder:
NOTE! existing checkouts will need to run:
git submodule foreach git pull origin master
git submodule update

PR #315, opened 2015-11-07
Fixup and normalize input of Okta logs

gdestuynder:
See also: https://github.com/jeffbryner/MozDef/issues/312 for normalization

This removes the rewriting of the config file, fixup some nits (whitespaces) and normalize a few useful fields.

Only tested shortly on mozdefqa1 so far.

From my experience writing this, I would propose creating a new issue to normalize all plugins and inputs, as well as creating a mozdeflib for parsing/reading (and usage of mozdef_client for sending, or both) so that we always query the fields the same way/easier to upgrade if fields format change.

@ameihm0912 as well for review

### DIFF
--- a/cron/okta2mozdef.py
+++ b/cron/okta2mozdef.py
@@ -133,6 +133,8 @@ def main():
                 else:
                     logger.error('Okta event does not contain published date: {0}'.format(event))
             state.write_state_file()
+        else:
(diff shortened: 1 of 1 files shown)


ameihm0912:
@gdestuynder looks good to me.

PR #342, opened 2016-05-24
Corrected  Case in investigationAdd.html

Phrozyn:
Corrected Case of Investigation Summary Field


Phrozyn:
Corrected from "investigation Summary"
to 
"Investigation Summary"

Issue #353, opened 2016-06-30
Feature Request: Blocklist aging


Phrozyn:
It would be great to have aging implemented on the blocklist, so that addresses are not blocked indefinitely. Most attacks are carried out by breached systems, meaning they aren't malicious forever  (except for the rare instance that it could be)

PR #358, opened 2016-07-01
Update blocklist to only look at 7 days of attackers

pwnbus:
References a possible fix for https://github.com/mozilla/MozDef/issues/353

### DIFF
--- a/cron/createIPBlockList.py
+++ b/cron/createIPBlockList.py
@@ -52,7 +52,7 @@ def toUTC(suspectedDate, localTimeZone=None):
     utc = pytz.UTC
     objDate = None
     if localTimeZone is None:
-        localTimeZone=options.defaulttimezone    
+        localTimeZone=options.defaulttimezone
     if type(suspectedDate) in (str, unicode):
         objDate = parse(suspectedDate, fuzzy=True)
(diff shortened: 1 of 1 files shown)


Phrozyn:
+r - requested testing on mozdefqa1 before committing.

jeffbryner:
why do we only care about 7 days?

pwnbus:
@jeffbryner @Phrozyn and I talked about this, and thought maybe it's best to bring up during our next infosec meeting real quick to get people's thoughts?

jeffbryner:
OK

Phrozyn:
Notes taken from an offline discussion:

- We should be deliberate about the block time value
- Distinguish first offense/second offense?
- Possible signals: if the IP is from AWS => shorter block
- Incorporate into roadmap for MozDef

Phrozyn:
I had to raise the number of IPs blacklisted in MozDef due to going over 2000, then over 3000.
We should implement this for now, maybe change it to 3 months of aging instead of 7 days, until we have more time to really make something beautiful out of this.

pwnbus:
Modified to reflect 3 months

pwnbus:
Closing PR since this change is no longer needed.

PR #362, opened 2016-07-13
Create utilities for toUTC with tests

pwnbus:

### DIFF
--- a/lib/utilities.py
+++ b/lib/utilities.py
@@ -1,0 +1,25 @@
+from datetime import datetime
+from dateutil.parser import parse
+import pytz
+
+def toUTC(suspectedDate, localTimeZone='UTC'):
+    '''make a UTC date out of almost anything'''
+
+    utc=pytz.UTC
+    objDate=None
+
+    if type(suspectedDate) in (str,unicode):
+        objDate=parse(suspectedDate,fuzzy=True)
+    elif type(suspectedDate)==datetime:
+        objDate=suspectedDate
+
+    if objDate.tzinfo is None:
+        objDate=pytz.timezone(localTimeZone).localize(objDate)
+        objDate=utc.normalize(objDate)
+    else:
+        objDate=utc.normalize(objDate)
+
+    if objDate is not None:
+        objDate=utc.normalize(objDate)
+
+    return objDate
--- a/lib/utilities.py
+++ b/lib/utilities.py
@@ -8,18 +8,18 @@ def toUTC(suspectedDate, localTimeZone='UTC'):
     utc=pytz.UTC
     objDate=None
 
-    if type(suspectedDate) in (str,unicode):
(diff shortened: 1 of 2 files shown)


lib/utilities.py
```diff
@@ -0,0 +1,25 @@
+from datetime import datetime
+from dateutil.parser import parse
+import pytz
+
+def toUTC(suspectedDate, localTimeZone='UTC'):
```
pwnbus: A couple notes here:

1. I decided to use UTC as the default timezone, I couldn't figure out how to grab the local timezone using tzlocal that didn't cause stuff to break, but the assumption I had was to assume we wanted UTC? I will probably circle back to this to see if I can use tzlocal.

2. I didn't move any of the cron scripts or anything that have this method redefined yet, I'm more or less looking for reviewers to see if this is how we want to tackle this.

pwnbus:
PLEASE DONT MERGE YET, just curious about reviewers, then I'll work on moving all callers of toUTC to use this utility function.

tests/test_utilities.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), '../lib'))
+from utilities import toUTC
+
+from datetime import datetime
+from dateutil.parser import parse
+
+class TestToUTC():
+
+  def result_is_datetime(self, result):
```
pwnbus: I need to switch from 2 spaces to 4 for this file....

lib/utilities.py
pwnbus: Although, now that I think about #1, do we want to have the timezone be dynamic? Maybe we should make a hard rule that any timestamps in mozdef, will always try and be UTC?

lib/utilities.py
```diff
@@ -0,0 +1,25 @@
...
+    '''make a UTC date out of almost anything'''
+
+    utc=pytz.UTC
+    objDate=None
+
+    if type(suspectedDate) == datetime:
+        objDate = suspectedDate
+    elif str(suspectedDate).isdigit():
+        # epoch? but seconds/milliseconds/nanoseconds (lookin at you heka)
+        epochDivisor = int(str(1) + '0'*(len(str(suspectedDate)) % 10))
```
pwnbus: I removed a couple methods to detect if a variable is a digit, and to return the number of digits. I believe we don't need to support numbers like 1+2i do we?

Phrozyn:
I can't answer that, I do know there are some timestamps with a date and time with the offset added at the end, I don't know if this is transforming those dates or if those are somehow transformed by the shipper or something else or if they are kept as a separate field.

pwnbus:
Closing this PR due to changes coming down the pipe that remove the need for this patch.

PR #364, opened 2016-07-14
Modified wording of Plugin Registration paragraph in code.rst

Phrozyn:
small modification to make it a little simpler.

### DIFF
--- a/docs/source/code.rst
+++ b/docs/source/code.rst
@@ -49,8 +49,7 @@ to assign custom document types, set static document _id values, etc.
 Plugin Registration
 +++++++++++++++++++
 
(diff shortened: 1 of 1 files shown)


jeffbryner:
'cept this section of the docs is talking about the event processing plugins, rather than the alert plugins so it's correct to reference the location of esworker.py

Probably should be separate sections for event plugins, alert plugins and rest plugins since they operate the same way ( registering for things, receiving messages) but live in different spaces since they modify different parts of the system (events/alerts/rest calls).

Phrozyn:
acknowledged, @jeffbryner 
I'll work on this.

PR #367, opened 2016-07-25
Set timezone to UTC by default

pwnbus:

### DIFF
--- a/benchmarking/workers/json2Mozdef.py
+++ b/benchmarking/workers/json2Mozdef.py
@@ -78,7 +78,7 @@ if __name__ == '__main__':
         for i in range(0,10):
 
             print(i)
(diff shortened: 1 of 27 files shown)


gdestuynder:
randomish-comment:
while toUTC is good to convert dates, to set the default date the program run as I would  consider using os.environ['TZ']='UTC'
The reason for that is that it correctly detects timezone as its already doing what toUTC does and then some, and has been in use for a few decades

Here's an example of implementation that still uses toUTC() and is backward compatible:

https://github.com/gdestuynder/rra2json/blob/master/rra2json.py#L49

pwnbus:
Closing PR due to changes coming down the pipe that removes most of these changes

PR #368, opened 2016-07-25
New init script

Phrozyn:
This script is for the new alertplugins mule created by @mpurzynski


Phrozyn:
Going to re-request after I back out that Plugin registration change as that is going to be revised.

PR #371, opened 2016-07-26
Manual Installation Instruction Update

Phrozyn:
Added instructions to manual installation for alertplugins and updated service startup instructions.


docs/source/installation.rst
```diff
@@ -568,29 +568,54 @@ Manual Installation
...
+    2. /etc/init.d/elasticsearch start or systemctl start elasticsearch
 
-    $ uwsgi --socket /run/uwsgi/apps/rest.socket --wsgi-file $MOZDEF_PATH/rest/index.py --buffer-size 32768 --master --listen 100 --uid root --pp $MOZDEF_PATH/rest --chmod-socket --logto /var/log/mozdef/uwsgi.rest.log -H $PATH_TO_VENV
+       $ service elasticsearch start
 
-    $ cd $MOZDEF_PATH/mq && uwsgi --socket /run/uwsgi/apps/esworker.socket --mule=esworker.py --mule=esworker.py --buffer-size 32768 --master --listen 100 --uid root --pp $MOZDEF_PATH/mq --stats 127.0.0.1:9192  --logto /var/log/mozdef/uwsgi.esworker.log --master-fifo /run/uwsgi/apps/esworker.fifo -H $PATH_TO_VENV
+    3. /etc/init.d/nginx start or systemctl start nginx
 
-    $ cd $MOZDEF_PATH/meteor && meteor run
+       $ service nginx start
```
gdestuynder: @Phrozyn 
shouldnt it be:
3.  /etc/init.d/nginx start or systemctl start nginx
$ /etc/init.d/nginx start
$ systemctl start nginx

instead?
(mainly "service" is only available on some distros anyway like RH/CentOS, but also just to be consistent in the doc - theres another instance above for elastisearch)

Phrozyn: This documentation is expressly for centos :)
Anyone running a different distro should be familiar with their startup commands, the debian flavored installation documentation is being written, but hasn't been committed yet.

Phrozyn: Also, aside from that - that's not what was changing in that commit.

PR #372, opened 2016-08-04
Add support to import auth0 logs into mozdef

gdestuynder:
This add support for our auth0 POC

### DIFF
--- a/cron/auth02mozdef.py
+++ b/cron/auth02mozdef.py
@@ -1,0 +1,331 @@
+#!/usr/bin/env python
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2016 Mozilla Corporation
+# Author: gdestuynder@mozilla.com
+
+# Imports auth0.com logs into MozDef
+
+import hjson
+import sys
+import requests
+import mozdef_client as mozdef
+try:
+    import urllib.parse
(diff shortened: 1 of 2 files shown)


gdestuynder:
@pwnbus @Phrozyn @mpurzynski  if one of you could review :)

cron/auth02mozdef.py
```diff
@@ -0,0 +1,331 @@
...
+import mozdef_client as mozdef
+try:
+    import urllib.parse
+    quote_url = urllib.parse.quote
+except ImportError:
+    #Well hello there python2 user!
+    import urllib
+    quote_url = urllib.quote
+
+class DotDict(dict):
```
pwnbus: More of a in the future type of thing, but I think we should get to a point where we have one class per file. I don't think this is a blocker or anything for now

pwnbus:
r+

Phrozyn:
I'm sorry I didn't get to this in time! Looks good though!

cron/auth02mozdef.py
gdestuynder: that would mean each of these feature potentially lives in their directory, such as:
cron/auth02mozdef/__init__.py
cron/auth02mozdef/DotDict.py
etc.

Not really against it though since these are currently self-contained programs its hard to say how it should be done (else i'd want a standard "shell" for the main code and run things as class/methods and have test cases)

pwnbus: Agreed, I think eventually they could have certain functions/classes exposed to the cron jobs, or like you said, some sort of "shell" that you overwrite a "run" or "execute" function to perform the custom code.

PR #374, opened 2016-08-12
Removing reference to "source" field

gene1wood:
As this field is not actually used by any event generators, it's not a mandatory field


pwnbus:
r+

PR #387, opened 2016-11-23
Fixup auth0 script with requirements

pwnbus:
Signed-off-by: Brandon Myers <bmyers@mozilla.com>


pwnbus:
Closing PR since it doesn't actually fix the problem...

Issue #366, opened 2016-07-22
Incorporate single startup script for uwsgi services


Phrozyn:
The idea is to run uwsgi in emperor mode and start all uwsgi services with a single startup script.
Point of reference on this is here:

https://www.digitalocean.com/community/tutorials/how-to-serve-django-applications-with-uwsgi-and-nginx-on-ubuntu-14-04

Phrozyn:
Once the above is done, a single systemd script can be used to start mozdef uwsgi processes

Phrozyn:
I've added a single startup script for the mozdefalertplugins for now with commit 8781118
This is so that mozdefalertplugins can be started until this "master init" is completed.

Phrozyn:
all uwsgi services now have a startup script. This was resolved with merge #389
I took a slightly different route and may change this in the future.
At this time every uwsgi worker can be started/stopped/restarted individually
All of the mq workers can be started/stopped/restarted by issuing systemctl start/stop/restart mozdefmqw*

mozdefweb, alertplugins, loginput, and restapi are stopped/started/restarted by 

- systemctl start/stop/restart mozdefloginput
- systemctl start/stop/restart mozdefrestapi
- systemctl start/stop/restart mozdefweb
- systemctl start/stop/restart mozdefalertplugins

PR #389, opened 2016-11-29
Convert2systemd

Phrozyn:
I'd l ike to add this directory to the public mozdef repo.
all uwsgi are now in systemd, as well as mongo

mongo will pull it's configuration from a mongo.conf file which I will commit shortly.


Phrozyn:
mongod conf is in original commit!

pwnbus:
r+

Phrozyn:
closes #366

PR #391, opened 2016-12-19
Changes to meteor to allow login via a http Header instead of persona

jeffbryner:
Persona is deprecated, these changes allow mozdef UI to consume a http header (via in this case) to use as a login token from whatever upstream IAM/OIDC solution is chosen to authenticate.


pwnbus:
r+

PR #400, opened 2017-06-27
Improve logic on which messages will be filtered by the IRC bot

tristanweir:
Fixes the alert logic to ensure that if an IRC channel is specified, the message will be displayed by the bot

### DIFF
--- a/alerts/lib/alerttask.py
+++ b/alerts/lib/alerttask.py
@@ -206,7 +206,7 @@ class AlertTask(Task):
             alert['notify_mozdefbot'] = False
 
         # If an alert sets specific ircchannel, then we should probably always notify in mozdefbot
-        if 'ircchannel' in alert:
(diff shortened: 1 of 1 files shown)


alerts/lib/alerttask.py
```diff
@@ -206,7 +206,7 @@ def tagBotNotify(self, alert):
             alert['notify_mozdefbot'] = False
 
         # If an alert sets specific ircchannel, then we should probably always notify in mozdefbot
-        if 'ircchannel' in alert:
+        if 'ircchannel' in alert and ('ircchannel' != "" and 'ircchannel' != None):
```
yashmehrotra: This should be
`if 'ircchannel' in alert and alert['ircchannel'] != '' and alert['ircchannel'] != None):`

as you are checking that key's value in the alert dict

tristanweir: Fixed. Appreciate the suggestion.

PR #379, opened 2016-10-22
improvements to dockerfile

cglewis:
- update docs to match
- set default cmd so it can run in the background
- combine commands to reduce size (savings of almost 250mb)
- reorder commands to improve caching when building
- fix up whitespace

PR #401, opened 2017-07-03
Merge dockerfile with #379

pwnbus:


pwnbus:
There will be more changes to come, since this dockerfile uses outdated versions of ES and other services.

PR #402, opened 2017-07-05
Add travisci to project and update unit tests

pwnbus:


tests/rest/test_rest_index.py
```diff
@@ -66,10 +66,11 @@ def test_route_endpoints(self):
 
             json_resp.sort()
 
-            assert json_resp[1]['url'].endswith(":9090/index.html#/dashboard/elasticsearch/Example SSH Dashboard") is True
+            assert json_resp[1]['url'].endswith("/app/kibana#/dashboard/dashboard/Example SSH Dashboard") is True
```
Phrozyn: I don't think dashboard should be in here twice.

tests/rest/test_rest_index.py
```diff
@@ -66,10 +66,11 @@ def test_route_endpoints(self):
...
             assert json_resp[1]['name'] == 'Example SSH Dashboard'
 
-            assert json_resp[0]['url'].endswith(":9090/index.html#/dashboard/elasticsearch/Example FTP Dashboard") is True
+
+            assert json_resp[0]['url'].endswith("/app/kibana#/dashboard/dashboard/Example FTP Dashboard") is True
```
Phrozyn: Same as previous comment, I don't think dashboard should be in the url twice.

tests/rest/test_rest_index.py
pwnbus: Good find, this was due to the rest config, I fixed that and the tests in https://github.com/mozilla/MozDef/pull/402/commits/a695692382f34cd60e707d2ddf7f871998aebcdc and https://github.com/mozilla/MozDef/pull/402/commits/bf465b0480138c5bb65eaa159ffd2d686bcfb0d8

PR #403, opened 2017-07-07
Add Python 3.6, just to see what happens

claudijd:
This is part troll and part "well, maybe?", let's find out!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - '2.7.11'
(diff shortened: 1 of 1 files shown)


claudijd:
well, it was worth a try.  Maybe when this is a priority, we can tackle this effort.

pwnbus:
We'd have to update our requirements file to use dependencies that support python3. Sad to see we didn't even get to running our unit tests to see.

PR #411, opened 2017-07-12
Update docker dependencies to working installation

pwnbus:
There will be more work on this front in the near future, but this is just to update the installation of the dependencies to a working state.


docker/conf/supervisor.conf
```diff
@@ -1,115 +1,75 @@
...
+priority=4
+command=/usr/sbin/nginx
 
+[program:meteor]
+user=mozdef
+priority=3
+command=bash -c 'source /opt/mozdef/envs/python/bin/activate && node bundle/main.js'
+directory=/opt/mozdef/envs/meteor/mozdef
+
+# # Inject sample data in MozDef ES directly.
```
pwnbus: This will eventually get readded, but I just wanted to update the installation of our dependencies, so this will be the next step.

PR #417, opened 2017-07-17
Update fxa alert and mq plugin

pwnbus:


jeffbryner:
details.action instead of details.path and the registration key should probably be lower case?

pwnbus:
@jeffbryner Fixed the details.path in https://github.com/mozilla/MozDef/pull/417/commits/9e0b613f3c28831a56ebf845423b0f0eb29a5931 and fixed the lowercase registration term in https://github.com/mozilla/MozDef/pull/417/commits/4b665d8771ac57e51ae7b47bc0a566cfed211f5c

PR #419, opened 2017-07-18
Added logic to allow the option of using meteor accounts for sign-in …

Phrozyn:
…on deployment.

### DIFF
--- a/meteor/app/lib/settings.js
+++ b/meteor/app/lib/settings.js
@@ -19,7 +19,7 @@ mozdef = {
   rootAPI: "https://localhost:8444",
   kibanaURL: "https://localhost:9443/app/kibana#",
   enableBlockIP: true,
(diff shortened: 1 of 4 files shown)


meteor/app/lib/settings.js
```diff
@@ -18,6 +19,7 @@ mozdef = {
   rootAPI: "https://localhost:8444",
   kibanaURL: "https://localhost:9443/app/kibana#",
   enableBlockIP: true,
-  enableClientAccountCreation: false
+  enableClientAccountCreation: false,
+  authenticationType: "meteor-password"
```
Phrozyn: enableClientAccountCreation is the toggle for enabling meteor's accounts-password.
authenticationType is for telling us which is set.
Need to determine which type to have set as default.

PR #420, opened 2017-07-20
Remove the beaver integration example, project seems to have died.

mpurzynski:


pwnbus:
I'm going to manually override the fact that the tests need to pass for this PR, since I did not have travis setup correctly to build on forked branch PRs. In the future, when a PR is created from a forked branch, the tests should run.

PR #430, opened 2017-08-08
Update contacts in documentation

pwnbus:


docs/source/conf.py
```diff
@@ -46,7 +46,7 @@
 
 # General information about the project.
 project = u'MozDef'
-copyright = u'2014, Jeff Bryner, Anthony Verez'
+copyright = u'2017, Mozilla'
```
Phrozyn: The date on the copyright establishes how far back the claim is established to be, 
It could look like copyright = 2014, 2017, Mozilla 
or just copyright = 2014, Mozilla

Either way

PR #433, opened 2017-08-15
Update geolite db

pwnbus:


cron/update_geolite_db.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def save_db_data(save_path, db_data):
+    logger.debug("Saving db data to " + save_path)
+    if os.path.isfile(save_path):
+        logger.debug("Overwriting db data in " + str(save_path))
+    else:
+        logger.debug("Creating new db data file at " + str(save_path))
+    with open(save_path, "wb+") as text_file:
+        text_file.write(db_data)
```
pwnbus: We should write to a temp file, then move the temp file into place.

pwnbus: Added in https://github.com/mozilla/MozDef/pull/433/commits/53828fafd8b1a338eefb1a193a204b246af1eefe

PR #441, opened 2017-08-18
Add logo to favicon and main title bar

pwnbus:
This modifies the logo on the main page to the pre existing https://github.com/mozilla/MozDef/blob/master/meteor/public/images/logo.png . It also adds a favicon of the same logo.


pwnbus:
Another possible idea for a logo for these 2 locations would be http://blog.seanmartell.com/wp-content/uploads/2012/03/logo_0061_62.png

pwnbus:
An example of the header logo change https://imgur.com/a/QirTv

pwnbus:
Maybe we can use this logo (or whatever one we decide on), in the readthedocs pages as well

jeffbryner:
Neither of these are suitable for a logo. I'd recommend just using a generic mozilla one like this instead: 
https://www.mozilla.org/favicon.ico

pwnbus:
@jeffbryner That sounds good to me, I figured since there was already an image in the repo called 'logo.png', that this was the preferred logo to use.

jeffbryner:
Ahh.. yeah, old cruft. That and this https://github.com/mozilla/MozDef/blob/master/meteor/public/images/concrete.jpg can probably be removed.

Phrozyn:
I'd like to drop the old mozilla logo in the grey tab on the upper right, and add this instead:
https://www.mozilla.org/media/img/mozorg/mozilla-256.4720741d4108.jpg

Though we can keep the grey tab (I can photoshop this into a vector graphic)) and just add the new mozilla logo there.

Due to the rebranding of projects, we really can't have a mozdef logo, everything is supposed to conform to a particular design as defined here: https://mozilla.ninja/logosystem
There is a favicon example as well which is just the M as @jeffbryner pointed out earlier.

pwnbus:
@Phrozyn @jeffbryner I updated the favicon with the mozilla.org one that you pointed towards. 

I've also removed the firefox logo image on the header on the left (next to the MozDef 'home' button), but kept the mozilla tab image on the right. I figured we would address updating the mozilla tab as part of a larger effort to update the look and feel.

Thoughts?

jeffbryner:
I think we are at the initiative phase of the brand guidelines: 
https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=61249802#BrandandIdentityGuide(orHowtoSuccessfullyBrandYourEarlyStageProject)-InitiativeWordmarkLockupExamples

and as such should endevour to either match those styles or replace the firefox graphic with the dino graphic that is non-initiative specific but generally approved for general graphic use.

pwnbus:
@jeffbryner Originally this was meant as a quick update of some of the images in mozdef (including adding a favicon), but this is turning into a lot larger work item.

I'm gonna close this PR and we can start the initiative phase of rebranding sometime.

PR #444, opened 2017-08-21
Add a new helper lambda to unit tests. Create timestamp based on now …

mpurzynski:
…with some fields modified.


pwnbus:
Can we close this PR in favor of https://github.com/mozilla/MozDef/pull/445 ? If you were trying to create a PR for the unit test functions, that's fine, then my comments from the 445 PR are relevant here.

Whichever you decide is fine with me!

PR #446, opened 2017-08-22
Add cron script to verify required fields on events

pwnbus:


jeffbryner:
This will likely report all cloudtrail events as errant ;-]

pwnbus:
@jeffbryner Yeah, I don't think we're gonna enable it in prod yet, but more just use it as reporting for now that we can start to tackle the larger issues, and work our way down. Hopefully eventually we'll be able to enable it in prod, but right now it'll be noisy every time it runs.

cron/verify_event_fields.conf
```diff
@@ -0,0 +1,3 @@
+[options]
+esservers = http://localhost:9200
+required_fields=category,hostname,processid,processname,severity,source,summary,tags,timestamp,receivedtimestamp
```
pwnbus: Add utctimestamp to this

PR #453, opened 2017-08-25
add gcc-c++ to avoid missing g++ during bcrypt build

jeffbryner:
Just tried building and was missing g++ when node tries to build bcrypt. Here's a patch to add it.


jeffbryner:
Hmm.. something is missing. Tried the build with multiple containers and error'd out also in bcrypt: 
gyp ERR! build error
gyp ERR! stack Error: not found: make

PR #458, opened 2017-09-01
Update installation.rst

CrazyLlama:
updated the docker install to be a tiny bit more verbose and not tell people to use sudo while making


pwnbus:
Merging even though our status checks are in limbo (I have a fix for this), but since this is only a documentation change, this doesn't carry any risk.

pwnbus:
Wahoo! Your first contribution!

PR #459, opened 2017-09-01
Remove unused packages for compose meteor dockerfile

pwnbus:


pwnbus:
This is a result of https://github.com/mozilla/MozDef/commit/77936b10cd18367229ad122c019eb51f3a7a252c adding unnecessary packages that meteor container doesn't need.

PR #461, opened 2017-09-05
Replace dots with underscores in filenames

Phrozyn:
Took the liberty of also adding underscores to some alerts names and worker.


mq/esworker_papertrail.py
```diff
@@ -547,4 +547,4 @@ def initConfig():
     lastPluginCheck = datetime.now()-timedelta(minutes=60)
     pluginList, lastPluginCheck = checkPlugins(pluginList, lastPluginCheck)
 
-    main()
+main()
```
pwnbus: We lost the indentation here, which needs to be there since we want to call 'main' only if we're running this python script.

mq/esworker_sqs.py
```diff
@@ -11,7 +11,7 @@
 # kombu's support for SQS is buggy
 # so this version uses boto
 # to read an SQS queue and put events into elastic search
-# in the same manner as esworker.py
+# in the same manner as esworker-eventtask.py
```
pwnbus: esworker has been renamed to "esworker_eventtask.py", not "esworker-eventtask.py"

mq/eventtask.ini
```diff
@@ -0,0 +1,25 @@
...
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+mule = esworker_eventtask.py
+pyargv = -c /opt/mozdef/envs/mozdef/mq/esworker_eventtask.conf
+log-syslog = syslog-worker
```
pwnbus: Should 'syslog-worker' be changed to reflect the new 'eventtask' naming scheme?

Phrozyn: oh! thanks for catching this! I had it changed, but guess I reverted when I backed out a commit.

mq/esworker_sqs.py
Phrozyn: yeah, typo, will fix

mq/esworker_papertrail.py
Phrozyn: ok, will fix

Phrozyn:
Closing this PR, replaced by #475

PR #466, opened 2017-09-06
Add fluentd container to consume syslog logs

pwnbus:
This will provide a 'syslog server' interface for sending events to mozdef.

### DIFF
--- a/docker/compose/docker-compose.yml
+++ b/docker/compose/docker-compose.yml
@@ -199,7 +199,7 @@ services:
     depends_on:
       - loginput
     ports:
-      - 514:514
+      - 514:514/udp
     networks:
       - default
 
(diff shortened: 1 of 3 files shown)


pwnbus:
I believe this requires more testing, it worked on tcp, but appears to not work on udp?

andrewkrug:
In compose you specifically have to bind UDP to get this to work with UDP output.  In my case I'm using 5140/udp to avoid conflicts with port < 1024 on my system and side step potential issues with SELinux on Fedora / RHEL.

`docker run -d --name fluentd \
  -p 5140:5140/udp -v /etc/fluentd:/fluentd/etc -e FLUENTD_CONF=fluentd.conf \
  fluentd-krug`

^^ My docker run command is there.

PR #470, opened 2017-09-08
Add alert tests verify rabbitmq

pwnbus:


pwnbus:
Since I needed some of the changes in https://github.com/mozilla/MozDef/pull/464  I merged those into here, so the changes that actually occur, are from https://github.com/mozilla/MozDef/pull/470/commits/e252eff6a6fd00615560c8a033e736fad9f079a2 on up.

PR #471, opened 2017-09-11
Add whois bot

pwnbus:


rest/index.py
```diff
@@ -562,7 +562,7 @@ def getWhois(ipaddress):
         whois = dict()
         ip = netaddr.IPNetwork(ipaddress)[0]
         if (not ip.is_loopback() and not ip.is_private() and not ip.is_reserved()):
-            whois = IPWhois(netaddr.IPNetwork(ipaddress)[0]).lookup()
+            whois = IPWhois(netaddr.IPNetwork(ipaddress)[0]).lookup_whois()
```
pwnbus: Using the old lookup method actually prints a deprecation warning. So this is to update our use of the library to use a nondeprecated function.

PR #475, opened 2017-09-19
Naming convention changes

Phrozyn:
- updated uWSGI workers that had dots in their names to use underscores instead
- updated some filenames for files that are templated that were camelcase to use underscores
- incorporated upstream changes that happened prior to this PR
- updated the syslog filter to incorporat naming changes
- updated systemd files to reflect templated conventions for standardization


Phrozyn:
Nevermind, this is not working the way it should. Closing.

PR #476, opened 2017-09-20
Add plugin set class to lib

pwnbus:
This PR creates a plugin set class, that will be used in the areas of mozdef that load plugins (alertplugins, mq).

### DIFF
--- a/lib/plugin_set.py
+++ b/lib/plugin_set.py
@@ -85,4 +85,8 @@ class PluginSet(object):
         return (message_copy, metadata)
 
     def send_message_to_plugin(self, plugin_class, message, metadata):
+        '''moving this logic to a separate function allows
+           different types of plugin_sets, such as alerts that might not care
+           about receiving metadata in its plugins
+        '''
         return plugin_class.onMessage(message, metadata)
--- a/lib/plugin_set.py
+++ b/lib/plugin_set.py
@@ -80,7 +80,7 @@ class PluginSet(object):
                     return (message, metadata)
         return (message, metadata)
 
-    def send_message_to_plugin(self, plugin_class, message, metadata):
+    def send_message_to_plugin(self, plugin_class, message, metadata=None):
         '''moving this logic to a separate function allows
            different types of plugin_sets, such as alerts that might not care
            about receiving metadata in its plugins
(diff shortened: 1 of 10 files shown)


lib/plugin_set.py
```diff
@@ -0,0 +1,92 @@
...
+           this function compares that registration list
+           to the current message and sends the message to plugins
+           in order
+        '''
+        if not isinstance(message, dict):
+            raise TypeError('event is type {0}, should be a dict'.format(type(message)))
+
+        # We deep copy the message in case the caller of this function wants
+        # to reference the original one (so we don't change it in here)
+        message_copy = copy.deepcopy(message)
```
pwnbus: This is different, and worth noting. The plugins have the ability to modify a message, so we need to copy the original message so the callers of the plugins, have access to it.

Thoughts?

pwnbus:
This is setting us up to specify or not specify a list of plugins to enable. This allows us to turn on or off certain plugins if we so choose.

lib/plugin_set.py
jeffbryner: plugins should be able to change a message such that the subsequent plugin gets the new message. This enables things like picking out or correcting an IP in one plugin, geo coding it in another.

pwnbus: Plugins still have the ability to do that, there are a couple examples in the tests. This is specifically if a plugin fails, that we won't have a half modified event, and have reference to the original in an esworker for example.

pwnbus: @jeffbryner The sendEventToPlugins method this was written based off of, returns the modified message, so this is how callers of the plugins, can handle the message after a plugin has modified it.

pwnbus: @jeffbryner I originally misspoke, there isn't a test for it specifically, but subsequent plugins will receive the updated message. I'll circle back though and will add a test for it.

jeffbryner: OK, as long as each plugin gets the output of the last plugin it sounds like you are doing the same thing as originally. If there is an error caught by the try/catch, where does that get reported so someone know there is a problem?

pwnbus: I decided to remove the deepcopy, so the expectation is that both the message that is passed to the function, and the value returned will be the modified message. This allows us to return None in a plugin, and that will indicate that we should drop this message.

PR #483, opened 2017-09-26
Create SSO dashboard alert plugin

pwnbus:
Depends on https://github.com/mozilla/MozDef/pull/476


alerts/plugins/sso_dashboard.py
```diff
@@ -0,0 +1,82 @@
...
+                logger.error("FAILED to open the configuration file" + str(config_file_path))
+        self.config = json_obj
+
+        # Set the alert ids based on alert name
+        for name, alert in self.config['alerts'].iteritems():
+            alert['alert_code'] = b2a_hex(name)
+
+        self.connect_db()
+
+        self.registration = '*'
```
jeffbryner: is this a wildcard registration?

pwnbus: @jeffbryner It is, the idea here was that our plugin will potentially want to run on all alerts, and that way, we can be more specific as to what fields we specifically want to match on, and throw to SSO. Ex: we want to match on a category of 'X', but not necessarily if any keys or values have 'X'.

jeffbryner: Maybe I'm not understanding, but it doesn't seem like it would need to run on all alerts, just geoIP? Unless I'm missing something, it also doesn't match the registration needing to be a list? https://github.com/mozilla/MozDef/blob/ed2f6e02c254a6eb8e60c375a126fb64c27e0a35/alerts/alertWorker.py#L32

pwnbus: What I was imaging was in the future, instead of having to modify the actual plugin code, we would only need to modify the configuration file in order to add more alerts that we will be sending. This way, the plugin will be more or less generic, and other folks can possibly use the plugin in their environment.

For the registration, I'm using the new plugin set class from https://github.com/mozilla/MozDef/pull/476  which does support both lists and strings as the value (https://github.com/mozilla/MozDef/pull/476/files#diff-028a2df2c7685ba77c2542250d06e793R72 is a specific test for it, which calls https://github.com/mozilla/MozDef/pull/476/files#diff-90debaea9cca010ebab44a2e8a9652b5 if you're curious).

jeffbryner: Thanks for the IRL explanation. I missed the part where * is added as a matching value as well. As we discussed, seems like a trade off of a refined registration system vs config file based matching that'll probably play out as we add more types of alerts.

alerts/plugins/sso_dashboard.py
```diff
@@ -0,0 +1,83 @@
...
+        self.dynamodb.put_item(Item=alert_record)
+
+    def onMessage(self, message):
+        for name, alert in self.config['alerts'].iteritems():
+            # we have a message that matches an alert we care about
+            if message['category'] == alert['category']:
+                full_email = message['details']['principal']
+                username = full_email.split('@')[0]
+                auth_full_username = self.config['auth_id_prefix'] + username
+                city = message['details']['locality_details']['city']
```
pwnbus: The original intention here was to keep this generic so other alerts could fit right in, but my thought was, we can make this more generic to fit whatever next alerts we care about.

alerts/plugins/sso_dashboard.json
```diff
@@ -0,0 +1,18 @@
+{
+  "alerts": {
+    "AlertGeomodel": { // The classname of the alert that we want to send to dashboard
+      "category": "geomodel", // This is how to verify if we match on an alert or not
+      "risk": "high", // low,medium,high values used to control display in dashboard
+      "summary": "Did you recently login from {0}, {1}?", // the text that is displayed to the user in the browser
```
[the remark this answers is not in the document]
jeffbryner: @tristanweir these design discussions happened here: https://github.com/mozilla-iam/sso-dashboard/issues/130 The todo here is to sync what we are sending with the UI elements that were implemented as the UI that was implemented is for low and medium risk alerts.

PR #484, opened 2017-09-27
Make toUTC identify local timezone dynamically

pwnbus:


pwnbus:
@jeffbryner I think I may have found a solution that can pull the local system timezone, so that we don't have to have 'UTC' hardcoded in the toUTC function. Thoughts?

tests/lib/utilities/test_toUTC.py
```diff
@@ -4,10 +4,28 @@
...
 sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../../lib'))
+import pytz
+
+import tzlocal
+
+
+def utc_timezone():
+    ''' This is a mock function, so when we run tests
+        we trick the system into thinking we're on UTC
+        no matter what the real timezone is
```
pwnbus: Instead of mocking the method, we could just simply require this test to run on a system that is using UTC. This will however, fail on other systems, but maybe it gives us higher confidence that this determination of the timezone is working correctly.

lib/utilities/toUTC.py
```diff
@@ -2,14 +2,15 @@
 from dateutil.parser import parse
 import pytz
 import math
+import tzlocal
+
+local_timezone = tzlocal.get_localzone()
```
pwnbus: I do question if there's a better way to do this. I didn't want that get_localzone() function called every time toUTC is called however...maybe a check for if it's None, set it?

jeffbryner: The only other way I know how to do it is via something like: 
https://github.com/jeffbryner/ldapChangeMonitor/blob/master/ldapChangeMonitor.py#L244

It would be interesting to profile.

Phrozyn:
This is waiting on further discussion per @pwnbus

pwnbus:
I did some benchmarking, and the faster of the implementations depends on if you import the library or not. Since we're only importing and running the function once, no matter how many times toUTC is actually called, this performance should have a very minimal affect.

jeffbryner:
Looks good, that's the faster of the two functions anyway (dateutil.tz.tzlocal seems super slower), but like you say it's only on startup rather than per event.

PR #485, opened 2017-09-27
Add ssh password auth violation alert

claudijd:
This adds warning alerts in cases where a public ssh endpoint is detected that allows password auth, which is a no-no.


claudijd:
This alert will trigger on events generated from pentest-automation...

https://github.com/mozilla/pentest-automation/blob/master/policy/external_ssh_policy.rb
https://github.com/mozilla/pentest-automation/blob/master/policy/external_ssh_policy.yaml

This is very similar to the open ports alert strategy, if you're looking for something to compare it to.

I set to warning for the time being to give it some time to run in prod.

PR #487, opened 2017-09-28
Create Event Class

pwnbus:
This PR adds an event class, that is currently used to make sure our required properties are present on each and every event. As a placeholder, we use the 'None' string.


lib/event.py
```diff
@@ -0,0 +1,40 @@
...
+
+class Event(dict):
+
+    # We set default vaules so that we can later
+    # create an alert around these, and know when events
+    # have to use defaults
+    DEFAULT_STRING = 'None'
+
+    def add_required_fields(self):
+        if 'receivedtimestamp' not in self:
```
pwnbus: These values are the required fields that we have listed in the docs. Do we want to make 'source', 'timestamp', and details, all required fields?

lib/elasticsearch_client.py
```diff
@@ -131,7 +132,11 @@ def save_alert(self, body, index='alerts', doc_type='alert', doc_id=None, bulk=F
         return self.save_object(index=index, doc_type=doc_type, body=body, doc_id=doc_id, bulk=bulk)
 
     def save_event(self, body, index='events', doc_type='event', doc_id=None, bulk=False):
-        return self.save_object(index=index, doc_type=doc_type, body=body, doc_id=doc_id, bulk=bulk)
+        if type(body) is str:
```
pwnbus: While we do this logic in save_object as well, there are cases in our code base (the mq/esworker_sns_sqs.py for example), where we call the save_event with the body being a string, and not json.

We can either do this parsing logic in save_event, or in the Event class __init__ function, but unfortunately I think we're going to have to have these checks in both save_object, and save_event.

lib/event.py
```diff
@@ -0,0 +1,40 @@
...
+        if 'mozdefhostname' not in self:
+            self['mozdefhostname'] = socket.gethostname()
+        if 'tags' not in self:
+            self['tags'] = []
+        if 'category' not in self:
+            self['category'] = self.DEFAULT_STRING
+        if 'hostname' not in self:
+            self['hostname'] = self.DEFAULT_STRING
+        if 'processid' not in self:
+            self['processid'] = self.DEFAULT_STRING
```
pwnbus: While logically, one would think processid is an int, we treat it in many areas of the codebase as a string, which is why I'm using the default string here.

lib/event.py
```diff
@@ -0,0 +1,40 @@
...
+        if 'processname' not in self:
+            self['processname'] = self.DEFAULT_STRING
+        if 'severity' not in self:
+            self['severity'] = self.DEFAULT_STRING
```
pwnbus: I wasn't 100% sure here if we wanted to set the severity to 'INFO', or a legitimate default value, or just use the 'None' placeholder here.

lib/event.py
jeffbryner: Yes, this is because it's easier to have ES default to use all strings, unanalyzed than to attempt to pick the correct types per field (lots of numeric options) and we never do queries by ranges of process ID anyway..

lib/event.py
jeffbryner: I'd say we want it to be a unique default so we can still tell if there are events that do not contain the required fields.

lib/event.py
mpurzynski: timestamp - is it always a copy of the UTC timestamp, or it can be anything that the source sets? Could it be a local (for the source) time in a local timezone?
source - if we can come up with some smart heuristics
details - source creates it, otherwise it's not present. I'm not sure we want to have it set to null, for example, because it's kind of like a pointer to a JSON dict

lib/event.py
mpurzynski: Let's keep it as string

lib/event.py
mpurzynski: UNDEFINED so we can query it later?

pwnbus: I don't think any events will have 'None' as the value, so we should be able to search for these events. I'm fine with switching the placeholder text to be whatever, if folks think 'None' isn't unique enough to search on.

jeffbryner: I'd suggest UNKNOWN in caps as it should be obvious and easy to search.

pwnbus:
Let's use 'UNKNOWN' as the place holder text. I'll update this.

lib/event.py
Phrozyn: I agree with @jeffbryner

PR #488, opened 2017-09-28
Add config option for rest and loginput for listening host address

pwnbus:
I originally modified rest and loginput to listen externally, but figured a config option would be best here.


pwnbus:
@Phrozyn I updated the config option name to be "listen_host", since I think that's a little more descriptive as to what that value is used for.

PR #495, opened 2017-10-04
Naming Convention and Logging Changes.

Phrozyn:
One commit rolling up naming convention changes and logging changes.
This should be committed when PR 341 from ansible-eis is merged.

- Removed camel case and dots from filenames in alerts and mq.
- Shortens log retention to 4 weeks - down from 17
- Removes mozilla specific logs from the config file - those will now be templated out using ansible-eis.

There may be stragglers as far as camel case goes, but I'm confident we can tackle those in future PR's.


Phrozyn:
Closing PR, seems bugged - there is no esworker.cloudtrail.py in the changes I'm merging.

PR #496, opened 2017-10-06
Extract action verb and add it along with readonly to the event

gene1wood:


gene1wood:
The notion that the first camel case word is the verb as well as which verbs are readonly comes from reviewing [a list of all AWS actions](https://gist.github.com/gene1wood/7c856601ab3c6ee92ce00326fab0cc3f)

PR #498, opened 2017-10-09
Migrate scanning alerts to the new data structure

mpurzynski:


mpurzynski:
Interesting idea, please open me a bug or a Trello card because changing the type field requires changes to fluentd configuration on 11 clusters (ansible++), changes to the BroFixup.py code and re-writing all the events that are already in MozDef and came from Bro.

jeffbryner:
looks like type is actually severity? i.e. info/notice/warning/etc?

mpurzynski:
Nope, this is the type of a Bro log - conn, ssl, x509, smb_files, notice (bro notice) and so on - over 20 of them.

jeffbryner:
Ahh, thanks. I've got a pet peeve against 'type' as well, mostly because there is usually a better description but also as phrozyn mentions _type is already a thing for ES and is really easy to misunderstand. Dunno if we need to go as far as "re-writing all the events that are already in MozDef and came from Bro" since they've already been alerted on, but we should settle on what to call it

http://words.bighugelabs.com/type
https://www.merriam-webster.com/thesaurus/type
variety?
subclass?
division?
branch?
family?
others? ;-]

mpurzynski:
https://github.com/mozilla/MozDef/blob/master/mq/plugins/broFixup.py <-- type is here, so we will need to change that plugin, unit tests, ansible fluentd configuration and that leaves us with inconsistent state of Bro events i.e. using the old field name

pwnbus:
Closing this PR since @mpurzynski implemented the updated alerts as generic_alerts.

PR #511, opened 2017-10-20
new cron to update aliases, removed events alias update from rotateIndexes.py

Phrozyn:
- removed events alias rotation from rotateIndexes.py
- added cron to rotate events alias only

Tested on development mozdef host successfully.

events-previous events-20171020 - - - 
alerts          alerts-201710   - - - 
alerts-previous alerts-201707   - - - 
events          events-20171021 - - -

### DIFF
--- a/cron/rollover_index_aliases.sh
+++ b/cron/rollover_index_aliases.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+# This is a temporary fix only
+
+if ! [ -x "$(command -v curl)" ]; then
+  exit 1
+fi
+
+alias_exists=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/aliases | grep "events\s"`
+events_index=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/events | grep "events\s"`
+rotate_alias=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/events | grep "events\s"`
+index_current_date=`date +%Y%m%d`
+index_new_date=$((index_current_date+1))
+index_old_date=$((index_current_date-1))
+successful_connect=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200`
+
+if [[ "${successful_connect}" == *"name"* ]]; then
+  printf "We connected successfully with curl.\n"
+else
+  exit 1
+fi
+
+if [[ "${alias_exists}" == "events "* ]]; then
+  printf "Event alias exists.\n"
+else
+  if [[ "${events_index}" == "events "* ]]; then
+    exit 1
+  fi
+fi
+
+printf "Rolling over events alias."
+curl -XPOST mozdef.private.scl3.mozilla.com:9200/_aliases -d' { "actions": [{ "add" : { "index" : "events-'${index_new_date}'", "alias" : "events" } }, { "remove" : { "index" : "events-'${index_current_date}'", "alias" : "events" } } ]}'
+
+printf "Rolling over events-previous alias."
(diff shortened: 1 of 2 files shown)


cron/rollover_index_aliases.sh
```diff
@@ -0,0 +1,34 @@
...
+
+alias_exists=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/aliases | grep "events\s"`
+events_index=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/events | grep "events\s"`
+rotate_alias=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200/_cat/events | grep "events\s"`
+index_current_date=`date +%Y%m%d`
+index_new_date=$((index_current_date+1))
+index_old_date=$((index_current_date-1))
+successful_connect=`curl -s -XGET mozdef.private.scl3.mozilla.com:9200`
+
+if [[ "${successful_connect}" == *"name"* ]]; then
```
gdestuynder: probably should use:
```
if [[ $? -eq 0 ]]; then
...
```
because it's more reliable and probably an important check
alternative:
```
curl -s -XGET mozdef.private.scl3.mozilla.com:9200` || {
   echo "curl connection failed"
   exit 127
}
```

gdestuynder: but otherwise, really, you want that for each command above instead (so that you know all commands worked)

cron/rollover_index_aliases.sh
```diff
@@ -0,0 +1,34 @@
...
+if [[ "${alias_exists}" == "events "* ]]; then
+  printf "Event alias exists.\n"
+else
+  if [[ "${events_index}" == "events "* ]]; then
+    exit 1
+  fi
+fi
+
+printf "Rolling over events alias."
+curl -XPOST mozdef.private.scl3.mozilla.com:9200/_aliases -d' { "actions": [{ "add" : { "index" : "events-'${index_new_date}'", "alias" : "events" } }, { "remove" : { "index" : "events-'${index_current_date}'", "alias" : "events" } } ]}'
```
gdestuynder: likewise id just add a quick error check:

```
curl -XPOST mozdef.private.scl3.mozilla.com:9200/_aliases -d' { "actions": [{ "add" : { "index" : "events-'${index_new_date}'", "alias" : "events" } }, { "remove" : { "index" : "events-'${index_current_date}'", "alias" : "events" } } ]}' || {
echo "ERROR: POST failed - failed to roll over events alias" 
exit 127
}
```

gdestuynder: (or with just warning if failure is acceptable)

cron/rollover_index_aliases.sh
```diff
@@ -0,0 +1,34 @@
...
+
+printf "Rolling over events-previous alias."
+curl -XPOST mozdef.private.scl3.mozilla.com:9200/_aliases -d' { "actions": [{ "add" : { "index" : "events-'${index_current_date}'", "alias" : "events-previous" } }, { "remove" : { "index" : "events-'${index_old_date}'", "alias" : "events-previous" } }, { "remove" : { "index" : "events-'${index_old_date}'", "alias" : "events-previous" } } ]}'
```
gdestuynder: same

jeffbryner:
This seems to replace python with a shell script that just does the same exact thing? 
https://github.com/elastic/elasticsearch-py/blob/3be98b0a09e6d0820efdc20966f086ea6b8b3396/elasticsearch/client/indices.py#L325

Phrozyn:
I'm going to close this PR, as I think this can be handled with what we have, just using a different call to the ES client than what we are using now.

Or we'll just use curator and not need any modifications to this at all.

PR #513, opened 2017-10-23
Resolve alias rollover

Phrozyn:


lib/elasticsearch_client.py
```diff
@@ -56,11 +56,21 @@ def create_index(self, index_name, ignore_fail=False, mapping=None):
...
+#            self.es_connection.indices.delete_alias(index='*', name=alias_name)
+#
+#        self.es_connection.indices.put_alias(index=index_name, name=alias_name)
+
+    def update_alias(self, oldindex, index, alias):
+        actions = None
+        if self.es_connection.indices.exists_alias(oldindex, alias):
+        # Add and remove aliases.
+            actions = [
+                {'remove': {'index': oldindex, 'alias': alias}},
```
jeffbryner: Might consider always having the remove match * for oldindex, since all we want is an atomic set the alias to be X. 

i.e. we don't really care what it is now, but we know what we want it to be, so we don't have to worry about what it currently may be.

cron/rotateIndexes.py
```diff
@@ -81,14 +84,19 @@ def esRotateIndexes():
                         logger.debug('Creating %s index' % newindex)
                         es.create_index(newindex)
                     # set aliases: events to events-YYYYMMDD
-                    # and events-previous to events-YYYYMMDD-1 for example
+                    # and events-previous to events-YYYYMMDD-1
                     logger.debug('Setting {0} alias to index: {1}'.format(index, newindex))
-                    es.create_alias(index, newindex)
+                    alias = index
+                    index = newindex
+                    es.update_alias(oldindex, index, alias)
```
tristanweir: Based on the assignment statements, we are functionally passing es.update_alias(oldindex, newindex, index). I get clarifying alias = index, since we are basically passing the date string (eg events-YYYYMMDD) but why the other reassignment?

cron/rotateIndexes.py
```diff
@@ -65,9 +66,11 @@ def esRotateIndexes():
             options.dobackup, options.rotation, options.pruning):
             try:
                 if rotation != 'none':
+                    previndex = index
                     oldindex = index
```
tristanweir: My guess is that we are going to lose the thread on the distinction between previndex & oldindex after a while. We should either comment clearly or refactor to have todayindex, yesterdayindex, twodaysagoindex

cron/rotateIndexes.py
```diff
@@ -81,14 +84,19 @@ def esRotateIndexes():
...
+                    # and events-previous to events-YYYYMMDD-1
                     logger.debug('Setting {0} alias to index: {1}'.format(index, newindex))
-                    es.create_alias(index, newindex)
+                    alias = index
+                    index = newindex
+                    es.update_alias(oldindex, index, alias)
                     if oldindex in indices:
-                        logger.debug('Setting {0}-previous alias to index: {1}'.format(index, oldindex))
-                        es.create_alias('%s-previous' % index, oldindex)
+                        alias = 'events-previous'
```
tristanweir: Does it meet expectations to hardcode in "events-previous" vs "events-YYYYMMDD-1"

cron/rotateIndexes.py
Phrozyn: In the abstraction layer we have 3 variables.
However, we have two aliases being used: events, events-previous

I changed the abstraction layer to only take in index, oldindex, and alias.
I couldn't reuse that function if it was taking in oldindex, newindex, previndex, alias, index, etc.

So I separated by an if statement the two calls (one calls for "events" alias update, and the other for "events-previous" update.

The casting is to populate what the abstraction layer expects with the correct values.
I do think this can be done more elegantly, but I think that would require a rewrite.

cron/rotateIndexes.py
Phrozyn: Definitely agreed, I'll add comments in a little bit.

lib/elasticsearch_client.py
Phrozyn: That makes sense, I always try to be as explicit as possible, but am ok with not being explicit where it's not needed.

cron/rotateIndexes.py
Phrozyn: That's actually an artifact I should have removed, we plugged in the events-previous into the config file, which I forgot to PR as well. 

I will resolve this. Thanks for catching it!

cron/rotateIndexes.py
Phrozyn: Done

lib/elasticsearch_client.py
Phrozyn: Done!

cron/rotateIndexes.py
Phrozyn: Done!

cron/rotateIndexes.py
```diff
@@ -160,12 +170,16 @@ def initConfig():
         '20,0,0',
         options.configfile).split(',')
         )
-
     options.weekly_rotation_indices = list(getConfig(
         'weekly_rotation_indices',
         'events',
         options.configfile).split(',')
         )
+    options.previous_alias = list(getConfig(
```
jeffbryner: Is this expected to be a list? Seems like it would be a single value?

Phrozyn: Fixed! Thank you

Phrozyn:
Closing in favor of #519

jeffbryner:
Looks good to me!

PR #535, opened 2017-11-01
Update the types of auth0 messages

pwnbus:


cron/auth02mozdef.py
```diff
@@ -208,11 +208,11 @@ def debug(msg):
                 "level": 3 # Error
                 },
         'sd': {
-                "event": 'Success delegation',
+                "event": 'Success Delegation',
```
pwnbus: I've updated this minor change since I have a script that was comparing the summary strings from the web ui. The hope is this script will be used in the future to compare and update the message types.

PR #538, opened 2017-11-06
Docker improvements

pwnbus:


docker/Dockerfile
```diff
@@ -15,7 +15,7 @@ FROM centos:7
 LABEL maintainer="mozdef@mozilla.com"
 
 ENV NODE_VERSION 4.7.0
-ENV METEOR_VERSION 1.4.4.2
+ENV METEOR_VERSION 1.4.2.3
```
pwnbus: @jeffbryner You recently modified this, however we aren't actually running the web ui using that version. https://github.com/mozilla/MozDef/blob/master/meteor/.meteor/release is the file that determines what version of meteor we will use (if we have a more recent version of meteor and try to build, we actually will download the version specified in that file).

I decided to revert it back to what we're running. If we choose to update meteor, we'll have to modify that file and then test things on qa as well.

docker/Dockerfile
```diff
@@ -70,12 +70,6 @@ RUN \
   curl -sL -o /opt/mozdef/nodesource.rpm https://rpm.nodesource.com/pub_4.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
   rpm -i --nosignature --force /opt/mozdef/nodesource.rpm && \
   yum install -y nodejs-$NODE_VERSION && \
-  npm install source-map-support@0.4.2 \
```
pwnbus: I moved this list of dependencies and stored them in https://github.com/mozilla/MozDef/pull/538/files#diff-d7366c300fa0f1d70a1278125050d8d7R10, so that we only need to do "meteor npm install". If/when we need to update the node libraries we use, we'll only need to modify the file, and not change any of the deployment.

docker/Dockerfile
jeffbryner: Does it break something to update?

pwnbus: In order to update, we'd have to update node as well (in production too). Since this task requires a bit more qa'ing and hand holding for prod, I think it would be best to have it in a separate PR.

Phrozyn:
Won't merge till @jeffbryner has a chance to review.

Issue #536, opened 2017-11-02
Add IP to "unknown country" warning


edmorley:
#### ISSUE TYPE
 - Feature Idea

#### COMPONENT NAME
[alerts/plugins/sso_dashboard.py](https://github.com/mozilla/MozDef/blob/d89c0eb6d098cabd8869c3b94ecf52b5d7fe2fd8/alerts/plugins/sso_dashboard.py)

#### CONFIGURATION
Mozilla's instance

#### OS / ENVIRONMENT
Mozilla's instance

#### DESCRIPTION
I filed an issue against the SSO dashboard repo (mozilla-iam/sso-dashboard#192) but was told the data came from MozDef, so re-filing here.

Upon visiting https://sso.mozilla.com/dashboard recently, I saw a "Did you recently login from Unknown, Norway?" warning. Now I very much doubt this is a real problem (and is likely just a geolocation inaccuracy), but to confirm either way I really need to know the IP address.

As such, please can the warning be modified to show the IP?

tristanweir:
We actually just planned dealing with "Unknown" cities this for our upcoming sprint, although the UX might be slightly different than surfacing the IP.

edmorley:
Hi! To clarify - in this case whilst the city was unknown, the country was also wrong, so fixing the city wouldn't help on it's own (unless intended fix is something different).

pwnbus:
@edmorley I decided to include the IP address in every message, at the end. We also filter out the city now, if it is unknown.

This means for your example, the modified summary would be 'Did you recently login from Norway (1.2.3.4)?' We will also be adding the date from which it happened to the beginning of the message, which should further help in identifying the legitimacy. 

Thanks for the recommendation!

pwnbus:
This has been pushed to production, so you should start to see the IP show up in the summary, as well as the city excluded if it is 'Unknown'. We haven't quite wrapped up the "showing the date in the summary" part yet, but hopefully that can get resolved soon.

edmorley:
Amazing! Thank you :-)

PR #541, opened 2017-11-09
Improve geomodel dashboard

pwnbus:
Originally, I thought we could create a generic plugin, that could handle which alerts to forward to the dashboard based solely off of a config. Unfortunately, we need some more flexibility than that, for example, modifying the summary if city == 'unknown'.

As a result, I decided to rename the generic plugin, to a specific geomodel plugin. In the future, we will need to create a new plugin for adding another alert to forward.


tests/alerts/plugins/test_dashboard_geomodel.py
```diff
@@ -50,12 +50,23 @@ def test_geoip_message_good(self):
...
+    def test_unknown_city_message(self):
+        message_dict = self.good_message_dict
+        message_dict['details']['locality_details']['city'] = 'UNKNOWN'
+        assert self.test_result_record is None
+        result_message = self.plugin.onMessage(message_dict)
+        assert result_message == self.good_message_dict
+        assert self.test_connect_called is True
+        result_db_entry = self.test_result_record
+        assert type(result_db_entry['alert_code']) is str
+        assert result_db_entry['summary'] == 'Did you recently login from United States (1.2.3.4)?'
```
pwnbus: I also modified the summary so that we stuff the IP address at the end of the message. This resolves https://github.com/mozilla/MozDef/issues/536

pwnbus:
@Phrozyn Agreed, I haven't created it yet because I was looking for a thumbs up on changing it from a generic plugin, to an alert specific dashboard plugin.

Phrozyn:
I approve! 
I think we can likely improve on this in the future using the generic idea in an abstracted framework possibly. We can talk more on that another time.

Phrozyn:
Please add the ansible-eis PR when you get a chance before merging this!

PR #545, opened 2017-11-12
Docker nginx redirect fix

Phrozyn:
Since we don't have ssl in the docker containers, removing redirect to ssl port as this sometimes breaks things (like dashboards)


Phrozyn:
Closing as this pulled in an additional change  that should not go with the nginx redirect. Will resubmit.

PR #548, opened 2017-11-13
Modifying rest index test to reflect nature of actual kibana links

Phrozyn:
In Kibana link URL's  spaces are really dashes


Phrozyn:
Adding to 547

PR #551, opened 2017-11-14
Add filterlog firewall mq plugin

pwnbus:
This allows us to parse filterlog type firewall message. For example, those coming from pfSense.


pwnbus:
This could use support for ICMP type messages.

PR #556, opened 2017-11-21
Add custom logout url for external header auth

pwnbus:


pwnbus:
I still need to verify this in qa.

pwnbus:
Verified this is working as expected in QA. All set to review.

PR #567, opened 2017-12-07
Correcting case of apiversion in defaultMappingTemplate.json

Phrozyn:


pwnbus:
@Phrozyn Can you resolve the conflict please?

Phrozyn:
I feel like the cloudtrail commit will need reworked due to the many items that it currently isn't handling which is causing the field type exceptions. Am I misunderstanding?

PR #569, opened 2017-12-08
Modify logger output for prune indexes

pwnbus:


Phrozyn:
Just curious why we are changing this to debug, I've not heard anything in regards to there being a problem with the output we have now.

pwnbus:
@Phrozyn Changing this to debug means that we will no longer get the cron email when this task completed successfully.

Phrozyn:
Makes sense! +R

PR #571, opened 2017-12-08
Revert cloudtrail worker, alert and mapping

pwnbus:


config/defaultMappingTemplate.json
```diff
@@ -156,6 +151,11 @@
             "format" : "dateOptionalTime",
             "type" : "date"
           },
+          "apiVersion" : {
+          "type" : "string",
```
Phrozyn: there should be a two space indentation after the parent key name (s= space):
              "apiVersion" : {
              ss"type" : "string",
              ss"index" : "not_analyzed",
              ss"doc_values" : true
              },

PR #575, opened 2017-12-18
Add feedback alert to handle feedback events

pwnbus:
Ready to merge, will sync up with IAM team when they're ready.


alerts/feedback_events.py
```diff
@@ -0,0 +1,40 @@
...
+from query_models import SearchQuery, TermMatch
+
+
+class AlertFeedbackEvents(AlertTask):
+    def main(self):
+        self._config = self.parse_json_alert_config('feedback_events.json')
+        search_query = SearchQuery(minutes=30)
+
+        search_query.add_must([
+            TermMatch('_type', 'event'),
```
Phrozyn: I think we should remove the match on _type since we are moving away from using this as a term we match on to align with the migration.

pwnbus: Good point, fixed in https://github.com/mozilla/MozDef/pull/575/commits/167a3d6374b779cb9d33a60f1dc3c6f80faa61e6

PR #586, opened 2018-01-04
Remove leftover contributors newlines and remove remaining contributors in files

pwnbus:


benchmarking/es/insert_simple.js
```diff
@@ -3,8 +3,6 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 // Copyright (c) 2014 Mozilla Corporation
 //
```
Phrozyn: Extra comment line here that should be removed to be consistent with the rest of the commit

benchmarking/es/search_all_fulltext.js
```diff
@@ -3,8 +3,6 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 // Copyright (c) 2014 Mozilla Corporation
 //
```
Phrozyn: Extra comment line here that should be removed to be consistent with the rest of the commit

PR #589, opened 2018-01-17
Clarify usage doc with receivedtimestamp purpose

gene1wood:


pwnbus:
Thanks for the patch!

PR #593, opened 2018-01-29
A new and better version of brofixup for syslog-ng plus some tiny cle…

mpurzynski:
…anups

### DIFF
--- a/mq/plugins/broFixup.py
+++ b/mq/plugins/broFixup.py
@@ -51,7 +51,7 @@ class message(object):
         and sets the doc_type
         '''
 
-        self.registration = ['bro', 'nsm', '*', 'duration']
+        self.registration = ['bro']
         self.priority = 5
         try:
             self.mozdefhostname = u'{0}'.format(node())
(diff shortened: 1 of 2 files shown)


mq/plugins/broFixup.py
```diff
@@ -2,8 +2,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # Copyright (c) 2017 Mozilla Corporation
+#
+# Contributors:
```
pwnbus: We no longer keep contributor lists in each file, so you'll want to remove lines 5 - 9

mq/plugins/broFixup.py
```diff
@@ -45,7 +51,7 @@ def __init__(self):
         and sets the doc_type
         '''
 
-        self.registration = ['bro', 'nsm']
+        self.registration = ['bro', 'nsm', '*', 'duration']
```
pwnbus: Do we really want to match on *? (Meaning, every event?)

mq/plugins/broFixup.py
gdestuynder: you might want to just add a contributor file at the root (common practice since its more discoverable than commits and make people happy to be recognized)

pwnbus: I'd disagree, I think the list of contributors by commit is sufficient here, as it stays updated automatically.

PR #594, opened 2018-01-31
Rotate indexes update (draft)

Phrozyn:


cron/rotateIndexes.py
```diff
@@ -72,6 +72,14 @@ def esRotateIndexes():
...
                             continue
                     if newindex not in indices:
+                        if 'alerts' in newindex:
+                            logger.debug('Creating %s index with single shard' % newindex)
+                            index_config = {
+                              "settings": {
+                                  "number_of_shards": 1
+                              }
+                            }
+                            es.create_index(newindex, index_config)
```
pwnbus: You'll need an else statement here.

Or 'else' we'll be creating the alerts index, then will be trying to create the same index.

tests/unit_test_suite.py
```diff
@@ -106,11 +106,11 @@ def setup_elasticsearch(self):
...
 
-        self.es_client.create_index(self.event_index_name, mapping=mapping_str)
+        self.es_client.create_index(self.event_index_name, index_config=mapping_str)
         self.es_client.create_alias('events', self.event_index_name)
-        self.es_client.create_index(self.previous_event_index_name, mapping=mapping_str)
+        self.es_client.create_index(self.previous_event_index_name, index_config=mapping_str)
         self.es_client.create_alias('events-previous', self.previous_event_index_name)
-        self.es_client.create_index(self.alert_index_name, mapping=mapping_str)
+        self.es_client.create_index(self.alert_index_name, index_config=mapping_str)
         self.es_client.create_alias('alerts', self.alert_index_name)
```
pwnbus: You'll also need to modify https://github.com/mozilla/MozDef/blob/master/tests/lib/test_elasticsearch_client.py#L444 to use the 'index_config' parameter name.

Phrozyn: Thanks @pwnbus!

cron/rotateIndexes.py
Phrozyn: done

pwnbus: You'll want to swap these lines with each other https://github.com/mozilla/MozDef/blob/72bd27d6bae4a585d65e0ffd0de17f7679b424d3/cron/rotateIndexes.py#L82-L83 and fix the indentation so they are both lined up.

You'll also want to swap https://github.com/mozilla/MozDef/blob/72bd27d6bae4a585d65e0ffd0de17f7679b424d3/cron/rotateIndexes.py#L85-L86 as well. TL;DR: log before we perform the action, so if the action goes wrong, the log statement will be accurate.

Phrozyn: ahh ok, makes sense :)

PR #601, opened 2018-02-07
Remove type from alerts

Phrozyn:


tests/alerts/test_ssh_password_auth_violation.py
```diff
@@ -43,16 +43,6 @@ class TestAlertSSHPasswordAuthViolation(AlertTestSuite):
         )
     )
 
-    events = AlertTestSuite.create_events(default_event, 10)
```
pwnbus: We need to modify alerts/ssh_password_auth_violation.py to remove _type as well.

tests/alerts/test_ssh_key.py
```diff
@@ -69,15 +69,6 @@ class TestSSHKey(AlertTestSuite):
         )
     )
 
-    event = AlertTestSuite.create_event(default_event)
```
pwnbus: Since we replaced '_type' with category in the alert, we should have a negative test with a bad category.

tests/alerts/test_session_opened_sensitive_user.py
```diff
@@ -126,16 +126,6 @@ class TestSessionOpenedUser(AlertTestSuite):
         )
```
pwnbus: Do we need to remove lines like https://github.com/mozilla/MozDef/blob/master/tests/alerts/test_geomodel.py#L14  from all of our alert unit tests?

tests/alerts/test_session_opened_sensitive_user.py
```diff
@@ -126,16 +126,6 @@ class TestSessionOpenedUser(AlertTestSuite):
...
     )
```
pwnbus: I wasn't sure where to put it, but looks like we have a couple alerts that still use the _type field:

$ grep -r '_type' ./alerts/ (with some filtering)
./alerts//amo_failed_logins.py:            TermMatch('_type', 'addons'),
./alerts//cloudtrail_logging_disabled.py:            TermMatch('_type', 'cloudtrail'),
./alerts//ssh_password_auth_violation.py:            TermMatch('_type', 'event'),
./alerts//ssh_ioc.py:            TermMatch('_type', 'event'),
./alerts//host_scanner_alerts.py:            TermMatch('_type', 'cef'), 

I'm also not sure what to do about these entries, thoughts?
./alerts//lib/alerttask.py:                'documenttype': e['_type'],
./alerts//lib/alerttask.py:                    'type': alertResultES['_type'],
./alerts//lib/alerttask.py:                self.es.save_event(index=event['_index'], doc_type=event['_type'], body=event['_source'], doc_id=event['_id'])

Phrozyn: Once we split the indexes we can remove all remaining calls to _type - this PR is only for removing the _type from the search parameters that trigger events. Low hanging fruit so to speak.

tests/alerts/test_ssh_key.py
Phrozyn: Good call, I have submitted a commit to resolve.

tests/alerts/test_session_opened_sensitive_user.py
Phrozyn: Not yet, not until we actually split the indexes.

tests/alerts/test_ssh_password_auth_violation.py
Phrozyn: This particular alert never fires, it will be part of our discussion on Wed.

tests/alerts/test_session_opened_sensitive_user.py
Phrozyn: These have notes on them in the spreadsheet as to why they aren't modified, many of them don't trigger as they are.

Phrozyn: as for alerttask.py -  should be removed when we split the indexes.

tests/alerts/test_ssh_password_auth_violation.py
Phrozyn: Per our discussion, this  is ok if it never fires (this is good) and I have modified _type to be category instead.

alerts/confluence_shell.py
```diff
@@ -1,46 +0,0 @@
-#!/usr/bin/env python
```
Phrozyn: This is deprecated as this system is no longer on our network. thus the removal.

alerts/correlated_alerts.py
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env python
```
Phrozyn: This is deprecated as we are no longer using CrowdStrike for this.

alerts/host_scanner_alerts.py
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env python
```
Phrozyn: This is deprecated as this solution was modified and lives on in ssh_key.py, thus removing

Phrozyn:
Something just got really messed up in the PR lol - going to close and resubmit.

PR #602, opened 2018-02-08
Modify cron scripts to remove doc_type

pwnbus:


cron/verify_event_fields.py
```diff
@@ -23,18 +23,18 @@ def verify_events(options):
         search_query.add_must_not(ExistsMatch(required_field))
 
         # Exclude all events that are mozdef related health and stats
-        search_query.add_must_not(TermMatch('_type', 'mozdefstats'))
-        search_query.add_must_not(TermMatch('_type', 'mozdefhealth'))
+        search_query.add_must_not(TermMatch('category', 'stats'))
```
pwnbus: I wonder if we should name these 'stats' or 'mozdefstats' to be more specific, thoughts?

Phrozyn:
I'm not positive this should be merged and pushed yet until we actually split the indexes, as _type is relevant up to the point we split the indexes, removing it prematurely will cause our data to be indexed differently, and any dashboards /saved searches currently in use with _type defined will break.
Removing _type from search params is ok though, that won't break anything.

PR #610, opened 2018-02-13
First pass at updating default mapping for ES 5

pwnbus:


config/defaultMappingTemplate.json
```diff
@@ -7,117 +7,103 @@
...
           "source" : {
-            "type" : "string",
-            "index" : "not_analyzed",
+            "type" : "keyword",
             "doc_values" : true
           },
           "summary" : {
-            "type" : "string"
+            "type" : "text",
+            "fielddata": true
```
pwnbus: In order to get our aggregation tests working properly, we had to add the fielddata to true. Thoughts?

jeffbryner: Where to we aggregate on the summary field? Seems like a memory killer according to: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html

config/defaultMappingTemplate.json
```diff
@@ -7,117 +7,103 @@
...
         }, {
           "double_fields" : {
             "mapping" : {
-              "index" : "not_analyzed",
               "doc_values" : true,
-              "type" : "string"
+              "type" : "keyword"
             },
             "match" : "*",
             "match_mapping_type" : "double"
```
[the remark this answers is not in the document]
Phrozyn: We do not store any "double" field types

config/defaultMappingTemplate.json
Phrozyn: I really don't like the idea of enabling fielddata, We'll still be able to search on the summary if it's set to Text based on any token within the field, but I don't know that we do any aggregations now based on single words in summary only searches for strings within the summary.

Phrozyn: Per https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html they recommend using keyword field for sorting/aggregation  - not text fields. I can't imagine aggregating on summary would be very beneficial if it returned every token in the field in an aggregation (like returning new and york rather than new york)

config/defaultMappingTemplate.json
```diff
@@ -138,12 +123,10 @@
                 "type" : "boolean"
```
Phrozyn: We should probably talk about whether we want doc_values here or keep it as is. The question is would we ever want to sort/aggregate on a boolean field.

config/defaultMappingTemplate.json
pwnbus: @jeffbryner We're aggregating based on summary in a lot of our generic alerts.

jeffbryner: ES aggregation or python aggregation? ala: https://github.com/mozilla/MozDef/blob/master/alerts/lib/alerttask.py#L243

pwnbus: Ahh, you're right, we don't do any ES aggregation in the alerts, I forgot that the aggregation in the alerts is python based. I think I can fix that test to use another keyname and will remove the fielddata as well from the mapping.

jeffbryner: Yeah, IIRC we did it that way because the ES aggregation made it difficult to get to the actual records themselves and we only deal with small data sets; so it's easier to grab the raw record set and let python aggregate.

jeffbryner:
FYI: I ran this template through a 6x docker image and got complaints about mappings for anything other than: "possible values are [object, string, long, double, boolean, date, binary]", so for float, short, byte, etc.

pwnbus:
@jeffbryner Yeah, this PR isn't complete, as there are still a couple things that need to get figured out/fixed. Also, FYI this is for ES5, not ES6 (yet).

pwnbus:
Closing PR since https://github.com/mozilla/MozDef/pull/625 is the preferred one now

PR #625, opened 2018-03-01
ES5 mapping update

pwnbus:


pwnbus:
Closing due to separating ES5 from ES6 changes.

PR #632, opened 2018-03-06
Elasticsearch 5 Upgrade

pwnbus:
Changes relevant to the Elasticsearch 5 upgrade.


docker/compose/nginx/files/nginx.conf
```diff
@@ -59,43 +59,8 @@ http {
     listen              *:9090;
     access_log          /dev/null main;
 
-    location ~ (/app/kibana|/app/sense|/bundles/|/kibana4|/status|/plugins|/elasticsearch|/api/sense/proxy|/kibana|/kibana/goto) {
```
Phrozyn: I suppose all of this is not needed for docker, but does sense still work if we remove the path to it?

docker/conf/jvm.options
```diff
@@ -0,0 +1,117 @@
...
+# If documents were already indexed with unquoted fields in a previous version
+# of Elasticsearch, some operations may throw errors.
+#
+# WARNING: This option will be removed in Elasticsearch 6.0.0 and is provided
+# only for migration purposes.
+#-Delasticsearch.json.allow_unquoted_field_names=true
+
+
+# Mozdef Custom Settings #
+-Xms512m
```
Phrozyn: You have these custom settings, but these settings are duplicated on lines 22 and 23 setting Xms and Xms to 2g, while these set it to 512m.
Which ones are to be the standard for a docker container and we should comment out the ones that are not.

docker/conf/nginx.conf
```diff
@@ -60,43 +60,6 @@ http {
...
-    # Password protected end points
-    location ~ ^/.kibana/dashboard/.*$ {
-      proxy_pass http://127.0.0.1:9200;
-      proxy_read_timeout 90;
-    }
-    location ~ ^/.kibana/temp.*$ {
-      proxy_pass http://127.0.0.1:9200;
-      proxy_read_timeout 90;
-    }
+    proxy_pass http://127.0.0.1:5601;
```
Phrozyn: curious why this line is here, and not in  docker/compose/nginx/files/nginx.conf - just seems a bit duplicative?

docker/conf/jvm.options
pwnbus: I took the full default config file, and did our custom settings at the very end, so next time if we need to update, we know exactly what we changed (the stuff in the Mozdef Custom Settings) at the end.

docker/conf/nginx.conf
pwnbus: This file is specific for the docker container where everything runs in one container, the file you pointed to is used for docker with multiple containers.

docker/compose/nginx/files/nginx.conf
pwnbus: Sense is no longer present in kibana, they removed the plugin entirely. They offer the same functionality built into kibana.

Phrozyn: oh nice! that's great news :)

docker/conf/jvm.options
Phrozyn: I understand that, but what I'm saying is all four lines are live settings - I feel two of them should be commented out to prevent any conflict should something go awry.

docker/conf/nginx.conf
Phrozyn: ahh ok! I missed that. Thanks!

PR #635, opened 2018-03-06
Remove type mq plugins

pwnbus:


pwnbus:
Replaces https://github.com/mozilla/MozDef/pull/627 since we're breaking out ES5 from ES6 changes.

PR #638, opened 2018-03-07
Update default mapping es5

pwnbus:


pwnbus:
@jeffbryner After having a meeting about it, this is what we propose as the changes to the mapping file, thoughts?

jeffbryner:
According to: https://www.elastic.co/guide/en/elasticsearch/guide/current/_deep_dive_on_doc_values.html 

> "Doc values are enabled by default for all fields except analyzed strings. That means all numerics, geo_points, dates, IPs and not_analyzed strings.
 Because doc values are on by default, you have the option to aggregate and sort on most fields in your dataset. But what if you know you will never aggregate, sort or script on a certain field?"

> 

We don't sort or script, so the only reason to keep doc values enabled would be aggregations. Do we aggregate on all fields?

https://www.elastic.co/guide/en/elasticsearch/reference/current/norms.html
I think in the previous template, norms were also turned off since we don't score.

Keyword as a datatype makes sense since most fields are a single value except summary which is a phrase.

https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html
I think because we don't score we can also make use of different index options?

tristanweir:
> We don't sort or script, so the only reason to keep doc values enabled would be aggregations. Do we aggregate on all fields?

Our theory is that we don't want to hamstring ourselves by excluding any fields from doc_values unless we _know_ we won't aggregate on them. In other words, keep the default of being able to aggregate on everything, and exclude things sparingly. 

According to [ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/doc-values.html),

> you can disable doc values in order to save disk space

Disk is not traditionally a limiter in our environment, although it could be in other deployments. Ultimately, I'd rather keep the option to aggregate available as long as disk capacity is available.

jeffbryner:
Sure, but I doubt we are going to aggregate on all the cloudtrail fields for example. So if we are trying to avoid sparse indexes: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_why_are_mapping_types_being_removed
it would seem removing doc_values everywhere and adding only where we need them would be a way to help.

Phrozyn:
That link doesn't mention doc_values, I think our inverted indexes are going to have sparsity as it is mainly because we are "indexing" all fields. I don't think doc_values really has any bearing on that other than the "metadata" it's keeping for scoring/aggregation/sorting

Anyhow, since there are supposedly improvements in the next version, let's just get off 2.x with what we have (minimal changes) and see where that puts us. If we find we have problems we can look at refining how we classify and assign these features to our data before moving to 6.x

PR #643, opened 2018-03-09
Update rotate index cron script to create index with settings

pwnbus:


cron/correlateUserMacAddress.py
```diff
@@ -134,7 +134,7 @@ def main():
 
     es = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
     # create intelligence index if it's not already there
-    es.create_index('intelligence', ignore_fail=True)
+    es.create_index('intelligence')
```
pwnbus: This change was merged into master, so I cherry-picked it here since we're technically dependent on it.

lib/elasticsearch_client.py
```diff
@@ -48,7 +48,7 @@ def delete_index(self, index_name, ignore_fail=False):
     def get_indices(self):
         return self.es_connection.indices.stats()['indices'].keys()
 
-    def create_index(self, index_name, ignore_fail=False, index_config=None):
+    def create_index(self, index_name, index_config=None):
```
pwnbus: Same with this, this change was merged into master, so I cherry-picked it here since we're technically dependent on it.

pwnbus:
Once https://github.com/mozilla/MozDef/pull/638 is merged, I'll circle back and move the mapping json file to cron, and will fix up all of the references.

pwnbus:
FYI this is failing due to the mapping changes not getting merged yet, once those are merged, this PR unit tests will start to pass.

cron/backup.conf
```diff
@@ -7,3 +7,8 @@ backup_dobackup = 1,1,1,1,1,1,1,1
 backup_rotation = none,none,monthly,daily,none,none,none,none
 backup_pruning = 0,0,0,180,0,0,0,0
 weekly_rotation_indices = events
+refresh_interval = 1s
```
Phrozyn: This should be set to 30s 
With as much indexing as we do setting it to 1s would create way too much overhead for our cluster.

cron/rotateIndexes.py
```diff
@@ -170,6 +176,14 @@ def initConfig():
         options.configfile).split(',')
         )
 
+    default_mapping_location = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'defaultMappingTemplate.json')
+    options.default_mapping_file = getConfig('default_mapping_file', default_mapping_location, options.configfile)
+    options.refresh_interval = getConfig('refresh_interval', '1s', options.configfile)
```
Phrozyn: Should change the default refresh interval here too to 30s

cron/backup.conf
pwnbus: For our specific config, I believe this is the case. Since this is the default, I used that value here.

Phrozyn: ahh okies :) ty

PR #649, opened 2018-03-27
Changed python virtualenv locations in cron/*.sh

### LINKED ISSUE #421: Change python virtualenv to /opt/mozdef/envs/python
This means we need to modify each cron/*.sh file to point to the new python binary.

scriptonist:
Changed python virtualenv to /opt/mozdef/envs/python in `cron/*.sh` files
FIxes #421

### DIFF
--- a/cron/auth02mozdef.sh
+++ b/cron/auth02mozdef.sh
@@ -6,5 +6,5 @@
 # Author: gdestuynder@mozilla.com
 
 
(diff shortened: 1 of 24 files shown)


pwnbus:
Thanks for the submission, I'm going to point this PR at a virtualenv specific branch "virtualenv_path_change", instead of merging directly into master. There are other changes that we are preparing that go along with this which will also be merged into "virtualenv_path_change" branch.

PR #661, opened 2018-04-10
Improve toutc function to handle negative values

pwnbus:


jeffbryner:
This catches a negative string, but I don't think it'll work for a negative number? something like -86000

PR #662, opened 2018-04-10
Remove myo integration

pwnbus:


jeffbryner:
😢 but looks like no progress on https://github.com/thalmiclabs/myo.js/issues/1 so it's too much trouble to set up for the utility of it

PR #666, opened 2018-04-16
Add publickey to TermsMatch 

Phrozyn:
Currently we do not see any failed logins that use publickeys, we should be seeing these.
I expect the rate of fire for this alert to increase significantly as we are about to find out where possible script issues are in addition to users attempting to log into hosts where their key is not present.


Phrozyn:
This will not cause a significant increase in alerts as originally thought.

PR #669, opened 2018-04-17
Correcting category typo

Phrozyn:


Phrozyn:
I'm not sure how this alert ever got added since it's failing in an unrelated area to the change.

pwnbus:
The tests are failing as expected, because we changed the attributes of the alert. You need to modify https://github.com/mozilla/MozDef/blob/master/tests/alerts/test_old_events.py#L14

Phrozyn:
ahh makes sense, when I looked at the output, it looked like it was failing for something completely different.

PR #670, opened 2018-04-19
Work in Progress: Alert plugin to send email when duo bypass is detected

zsck:
This is a work-in-progress pull request that we can expect to add to.

For now, we have  an alert plugin that will send emails when a 'bypassused' alert is created.


pwnbus:
Closing PR since this was implemented in another method.

PR #683, opened 2018-05-08
Add key exists lib

pwnbus:


lib/utilities/key_exists.py
```diff
@@ -0,0 +1,17 @@
+
+def key_exists(search_key, inputed_dict):
```
pwnbus: Now instead of having multiple if statements when searching for 'details.<somekey>', we can use this utility function to make things cleaner.

PR #690, opened 2018-05-10
Cachemaint utility, closes #689

### LINKED ISSUE #689: Automate JVM/Cache maintenance
#### ISSUE TYPE
 - Feature

#### COMPONENT NAME
esCacheMaint.py

#### CONFIGURATION
esCacheMaint.conf

#### DESCRIPTION
ES can run high utilization of the JVM over time, usually as a result of having a series of indices retaining fielddata. Add (back) a utility to periodically check JVM utilization, check for fielddata cache on unused indices and if found, clear the cache using: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-clearcache.html

jeffbryner:


cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
...
+                logger.info('{0}: cpu {1}%  jvm {2}% load average: {3} recommending cache clear'.format(nodename,cpuusage,jvmused,loadaverage))
+                return True
+        return False
+    else:
+        logger.error(r)
+        return False
+
+def clearESCache():
+    es=esConnect(None)
+    indexes=es.indices.stats()['indices']
```
pwnbus: It's preferred to use our elasticsearch client to get this list https://github.com/mozilla/MozDef/blob/master/lib/elasticsearch_client.py#L48

cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
...
+    logger.addHandler(sh)
+
+def esConnect(conn):
+    '''open or re-open a connection to elastic search'''
+    return Elasticsearch((options.esservers))
+
+
+def isJVMMemoryHigh():
+    url = "http://{0}/_nodes/stats?pretty=true".format(random.choice(options.esservers))
+    #r=requests.get(url="http://mozdefes5.private.scl3.mozilla.com:9200/_nodes/stats?pretty=true")
```
pwnbus: Need to remove this.

cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import json
+import logging
+import pytz
```
pwnbus: This library isn't used so we can remove it

cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
...
+import random
+import requests
+import sys
+import time
```
pwnbus: This library isn't used so we can remove it

cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
...
+from configlib import getConfig, OptionParser
+from datetime import datetime, date, timedelta
+from dateutil.parser import parse
```
pwnbus: This library isn't used so we can remove it

cron/esCacheMaint.py
```diff
@@ -0,0 +1,121 @@
...
+from elasticsearch import Elasticsearch
+logger = logging.getLogger(sys.argv[0])
```
pwnbus: We also have a helper to create loggers, so you can simply "from utilities.logger import logger" and you no longer need this line, and the loggerTimeStamp and initLogger functions defined in this file.

Phrozyn:
I know I responded to this in email, but got to thinking about it later on. The reason the  clearing of cache didn't "help" when I ran it was due to not many people running complex searches. However, as Kibana is used more and more for complex searches clearing the cache could definitely help during those cycles.

PR #700, opened 2018-05-23
Fix XHR 302s with an access proxy. This fixes the page "hanging" after the access proxy tries to silently authenticate.

gdestuynder:
What would happen:

- XHR triggers while page is open
- User session has to be verified with an authentication provider by an
access proxy (Meteor is unaware of this)
- XHR gets a 302 reply to the authentication provider, but doesn't know
what to do with it.
- Additionally, CORS headers deny the 302 (rightfully so)
- All XHRs fail from now on, Mozdef webui stops working until you
manually refresh the page.

What happens with this patch:

- XHR triggers while page is open
- User session has to be verified with an authentication provider by an
access proxy (Meteor is unaware of this)
+ XHR reply is intercepted and checks if it's a 302 or 0 (CORS denial)
status
++ if so, the page is automatically refreshed and the user gets silently
reauthenticated as needed

### DIFF
--- a/meteor/app/client/mozdef.js
+++ b/meteor/app/client/mozdef.js
@@ -393,4 +393,37 @@ if (Meteor.isClient) {
     Meteor.logoutViaAccounts = function(callback) {
         return Accounts.logout(callback);
     };
+    // Intercepts all XHRs and reload the main browser window on redirect or request error (such as CORS denying access)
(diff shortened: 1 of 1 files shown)


gdestuynder:
Note: this code is currently live on https://mozdefqa2.private.scl3.mozilla.com/ with 10s refresh for testing.
Real world scenarios use 15min refresh

PR #712, opened 2018-06-06
Guard duty plugin

jeffbryner:
Plug in and updates to docdict class to allow for easier fixup of guard duty messages. This fixes dates to be isoformat and pulls IP addresses from some common locations

### DIFF
--- a/lib/utilities/dot_dict.py
+++ b/lib/utilities/dot_dict.py
@@ -17,3 +17,27 @@ class DotDict(dict):
             if hasattr(value, 'keys'):
                 value = DotDict(value)
             self[key] = value
+
+    def get(self, key):
+        """get to allow for dot string notation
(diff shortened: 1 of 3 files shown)


lib/utilities/dot_dict.py
```diff
@@ -17,3 +17,27 @@ def __init__(self, dct={}):
             if hasattr(value, 'keys'):
                 value = DotDict(value)
             self[key] = value
+
+    def get(self, key):
```
pwnbus: This class is somewhat unit tested, so is there any chance we can add a few tests for this function to https://github.com/mozilla/MozDef/blob/master/tests/lib/utilities/test_dot_dict.py

jeffbryner: Sure, added some that should exercise the get('string.string') a bit

PR #719, opened 2018-07-05
Add FQDN/DNS block page

claudijd:
I'm not very "webby" in the context of developing web-pages, though I do like to break them.  Here's an attempt at cloning something we've already built for IP/network block lists and porting to work for FQDNs (Fully Qualified Domain Names).  The idea being that if a SPOC needs to respond to say block a phishing domain, they could drop into the UI and take decisive action to "add it to the FQDN blocklist" and that could be consumed by other services (see bug 1454672), like the proxy servers, DNS servers, and maybe even AWS security groups to protect our users and our services from said bad FQDN.

I've also marked this a WIP (Work In Progress) to start with since I'm starting with just modding code, getting some initial review, and I've not actually even tested this to make sure it doesn't immediately blow up.

If you're interested in helping in any way, please feel free to do so in whatever way you feel comfortable.

### DIFF
--- a/cron/createFDQNBlockList.py
+++ b/cron/createFDQNBlockList.py
@@ -104,12 +104,12 @@ def main():
                     ]},
                 },
                 {"$project":{"address":1}},
-                {"$limit": options.iplimit}
+                {"$limit": options.fqdnlimit}
             ])
         FQDNList=[]
         for fqdn in fqdnCursor:
             # TODO: figure out what to do here
-            FQDNList.append(ip['address'])
+            FQDNList.append(fqdn['address'])
         # to text
         with open(options.outputfile, 'w') as outputfile:
             for fqdn in FQDNList:
@@ -121,7 +121,7 @@ def main():
 
 
     except ValueError as e:
-        logger.error("Exception %r generating IP block list" % e)
+        logger.error("Exception %r generating FQDN block list" % e)
 
 
 def initConfig():
@@ -152,7 +152,7 @@ def initConfig():
     options.expireage = getConfig('expireage',1,options.configfile)
 
     # Max FQDNs to emit
-    options.iplimit = getConfig('fqdnlimit', 1000, options.configfile)
+    options.fqdnlimit = getConfig('fqdnlimit', 1000, options.configfile)
 
     # AWS creds
     options.aws_access_key_id=getConfig('aws_access_key_id','',options.configfile)          #aws credentials to use to connect to mozilla_infosec_blocklist
--- a/cron/createFDQNBlockList.py
+++ b/cron/createFDQNBlockList.py
@@ -124,7 +124,7 @@ def initConfig():
     options.mongoport = getConfig('mongoport', 3001, options.configfile)
 
     # FQDN whitelist as a \n separted file of example.com or foo.bar.com style names
-    self.options.fqdn_whitelist_file = getConfig('fqdn_whitelist_file', '/dev/null', self.configfile)
+    options.fqdn_whitelist_file = getConfig('fqdn_whitelist_file', '/dev/null', options.configfile)
     options.fqdnwhitelist = parse_fqdn_whitelist(options.fqdn_whitelist_file)
 
     # Output File Name
--- a/cron/createFDQNBlockList.py
+++ b/cron/createFDQNBlockList.py
@@ -78,7 +78,6 @@ def main():
         client = MongoClient(options.mongohost, options.mongoport)
         mozdefdb = client.meteor
         fqdnblocklist = mozdefdb['fqdnblocklist']
-        attackers=mozdefdb['attackers']
         # ensure indexes
         fqdnblocklist.create_index([('dateExpiring',-1)])
 
@@ -148,9 +147,6 @@ def initConfig():
 
     # Category to choose
     options.category = getConfig('category', 'bruteforcer', options.configfile)
-
-    # Max days to look back for attackers
-    #options.attackerage = getConfig('attackerage',90,options.configfile)
 
     # Days after expiration that we purge an fqdnblocklist entry (from the ui, they don't end up in the export after expiring)
     options.expireage = getConfig('expireage',1,options.configfile)
--- a/cron/createFDQNBlockList.py
+++ b/cron/createFDQNBlockList.py
@@ -61,44 +61,6 @@ def isFQDN(fqdn):
     except:
         return False
 
-# pretty sure this isn't relevant in an FDQN context
-# TODO: be sure of this before removing fully
-# def aggregateAttackerIPs(attackers):
-#     iplist = []
-
-#     # Set the attacker age timestamp
-#     attackerage = datetime.now() - timedelta(days=options.attackerage)
-
-#     ips = attackers.aggregate([
-#         {"$sort": {"lastseentimestamp": -1}},
-#         {"$match": {"category": options.category}},
-#         {"$match": {"lastseentimestamp": {"$gte": attackerage}}},
-#         {"$match": {"indicators.ipv4address": {"$exists": True}}},
-#         {"$group": {"_id": {"ipv4address": "$indicators.ipv4address"}}},
-#         {"$unwind": "$_id.ipv4address"},
-#         {"$limit": options.iplimit}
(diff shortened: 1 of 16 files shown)


cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+
+def main():
+    logger.debug('starting')
+    logger.debug(options)
+    try:
+        # connect to mongo
+        client = MongoClient(options.mongohost, options.mongoport)
+        mozdefdb = client.meteor
+        fqdnblocklist = mozdefdb['fqdnblocklist']
+        attackers=mozdefdb['attackers']
```
jeffbryner: can remove since attackers don't have dns

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+                {"$sort": {"dateAdded": -1}},
+                {"$match": {"address": {"$exists": True}}},
+                {"$match":
+                    {"$or":[
+                        {"dateExpiring": {"$gte": datetime.utcnow()}},
+                        {"dateExpiring": {"$exists": False}},
+                    ]},
+                },
+                {"$project":{"address":1}},
+                {"$limit": options.iplimit}
```
jeffbryner: probably needs a new option. The option is just to limit any list to X entries.

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+            ])
+        FQDNList=[]
+        for fqdn in fqdnCursor:
+            # TODO: figure out what to do here
+            FQDNList.append(ip['address'])
```
jeffbryner: the way mongo works, since you've defined an item as 'fqdn' in fqdnCursor you'd reference the address as fqdn['address']

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+            for fqdn in FQDNList:
+                outputfile.write("{0}\n".format(fqdn))
+        outputfile.close()
+        # to s3?
+        if len(options.aws_bucket_name)>0:
+            s3_upload_file(options.outputfile, options.aws_bucket_name, options.aws_document_key_name)
+
+
+    except ValueError as e:
+        logger.error("Exception %r generating IP block list" % e)
```
jeffbryner: s/IP/DNS

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+    options.category = getConfig('category', 'bruteforcer', options.configfile)
+
+    # Max days to look back for attackers
+    #options.attackerage = getConfig('attackerage',90,options.configfile)
+
+    # Days after expiration that we purge an fqdnblocklist entry (from the ui, they don't end up in the export after expiring)
+    options.expireage = getConfig('expireage',1,options.configfile)
+
+    # Max FQDNs to emit
+    options.iplimit = getConfig('fqdnlimit', 1000, options.configfile)
```
jeffbryner: s/iplimit/fqdnlimit

meteor/app/client/blockFDQN.js
```diff
@@ -0,0 +1,63 @@
+/*
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at http://mozilla.org/MPL/2.0/.
+Copyright (c) 2014 Mozilla Corporation
+ */
+
+if (Meteor.isClient) {
+
+    Template.blockIPform.rendered = function() {
```
jeffbryner: The format is Template.<templatename>.event so for this it should be Template.blockFQDNform.rendered since you want it to autofill a block in the fqdn form.

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,248 @@
...
+        # Positive Example: foo.example.com (3 tokens, 'example' and 'com' == valid)
+        # Negative Example: com (1 token, 'com' == invalid)
+        if '.' in fqdn and len(ip.split('.'))>=2:
+            return True
+        else:
+            return False
+    except:
+        return False
+
+# pretty sure this isn't relevant in an FDQN context
```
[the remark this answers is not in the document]
claudijd: Removed

cron/createFDQNBlockList.py
claudijd: Removed

cron/createFDQNBlockList.py
claudijd: changed to fqdnlimit, but not sure if I need to worry about namespacing $limit

cron/createFDQNBlockList.py
claudijd: fixed

cron/createFDQNBlockList.py
claudijd: fixed, went with FQDN for consistency, but the entry point for the user will have FQDN/DNS in the UI, so it's not like a "what is FQDN?" type response.

cron/createFDQNBlockList.py
claudijd: fixed

meteor/app/client/blockFDQN.js
claudijd: fixed

jeffbryner:
I did some fixup on it and I think it's ready to merge.

claudijd:
Another question here, do I need to produce an FQDN white-list and upload in the same process as we do for Mozilla ranges or would we want that white-list to live closer to MozDef?

jeffbryner:
Good question, probably for another thread. Several ideas about automation, existing observatory inventory, etc come to mind.

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,182 @@
...
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import boto
+import boto.s3
+import logging
+import netaddr
```
pwnbus: We don't actually use this "netaddr" library in this file, can we remove it?

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,182 @@
...
+
+def isFQDN(fqdn):
+    try:
+        # We could resolve FQDNs here, but that could tip our hand and it's
+        # possible us investigating could trigger other alerts.  As such,
+        # validation will consist of making sure we have a dot noted string
+        #with two or more tokens
+        # Positive Example: example.com (2 tokens, 'example' and 'com' == valid)
+        # Positive Example: foo.example.com (3 tokens, 'example' and 'com' == valid)
+        # Negative Example: com (1 token, 'com' == invalid)
```
pwnbus: What about if we moved this function into lib, and then took the examples in the comments and made unit tests for them?

cron/createFDQNBlockList.py
```diff
@@ -0,0 +1,182 @@
...
+                                       options.configfile)
+    # syslog port
+    options.syslogport = getConfig('syslogport', 514, options.configfile)
+
+    # mongo instance
+    options.mongohost = getConfig('mongohost', 'localhost', options.configfile)
+    options.mongoport = getConfig('mongoport', 3001, options.configfile)
+
+    # FQDN whitelist as a \n separted file of example.com or foo.bar.com style names
+    self.options.fqdn_whitelist_file = getConfig('fqdn_whitelist_file', '/dev/null', self.configfile)
```
pwnbus: This line will error out, "self" is not defined.

rest/plugins/fqdnblocklist.py
```diff
@@ -0,0 +1,258 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import netaddr
```
pwnbus: Unused import

PR #723, opened 2018-07-12
Mozdefhealth index - initial review

Phrozyn:
I didn't change any of the other healthAndStatus crons as I would like a review of this, the index is never rotated (and for now shouldn't need to be until we see what kind of sizing it has after some time, my guess is that it won't get very big, and these are tiny docs.

The only change to the other crons needed is to add the index to the es.save, we'll let this cron handle the heavy lifting. If the other two  can't post for a second or two because an index is being created, it's not a problem.

Shards = 1, Replicas = 0, Changed to: Shards = 1, Replicas = 1


jeffbryner:
What problem are we solving for?

Phrozyn:
This is to remove non-security event data from the events index and reduces the mapping requirement for the index by a tiny margin.
Upon daily events index creation, this data is no longer dynamically created as it is statically mapped into it's own index.
This also extends the ability to add further metrics and time series the data without creating overhead by having to search through a large data set like events.

jeffbryner:
The organizing principle for indexes is retention period, rather than data type. i.e. things that only matter for a day are in events-YYYYMMDD. Things that need to be retained longer like alerts in monthly indexes. The type of data (they are all security events) doesn't matter so much or else we'd end up with dozens of indexes and exploding complexity. 

If we are worried about historical data increasing the size we can remove this: 
https://github.com/Phrozyn/MozDef/blob/6aad873903557083022c2bb1635aa27d011077a7/cron/healthAndStatus.py#L137

and just limit it to creating a static doc ID for current ala: 
https://github.com/Phrozyn/MozDef/blob/3362fad060734b88e889ee8be39d488d66cc3052/cron/healthAndStatus.py#L141

then you only have one document per host per day but I'm not sure it's worth even that change to save what's probably minimal space?

Phrozyn:
This is not an effort to save space, I apologize if that is how my response was interpreted.
I was speaking more to the advantage of having health data in it's own index for review of trending without creating overhead from having to search a larger data set for it.  This is to help with any troubleshooting  of health issues and provide trending so we can optimally plan for things before the problems present themselves. 
This is specifically for historical data on the health of the cluster and letting events be used for security work, while health data is used for ops work, and doesn't touch security event data.

Also to address the other assumption of splitting events up into further tiny indexes, that's not something we are planning on doing, although there is an argument for moving nsm data into it's own index (namely shard size so we  can prevent from having to add more servers to accommodate larger shards)

In addition to the above, the addition of the index_exists to the elasticsearch_client can optimize some of the other crons we have where we look for specific indexes upon rotation by eliminating the return of a large dataset, we can pick a specific index.

jeffbryner:
Ah, ok thanks. This is about data retention for health/status/event stats events, which makes sense. i.e. rather than keep 100s of millions of events online longer; keep smaller sets of data online longer for greater historical trending. 

FYI: https://github.com/mozilla/MozDef/blob/master/cron/eventStats.py#L87 should also get a PR then to allow it to post into the same index. It's the thing that summarizes counts of events per category every X minutes for historical trending of incoming events.

Phrozyn:
Yeah I plan on tackling that next, it'd be nice to see the trending for everything :)

Thanks, Jeff! Sorry my initial post wasn't more informative.

jeffbryner:
No worries, thanks for explaining the effort!

cron/mozdefStateMappingTemplate.json
```diff
@@ -0,0 +1,278 @@
+{
+    "order": 0,
```
pwnbus: Since most of this file isn't relevant to the mozdefstate index, can we instead just represent the options we care about (replicas/shards) in a dictionary object within the cron script?
Example:
```
mapping = {
    "settings": {
        "number_of_shards": 1,
         "number_of_replicas": 1
    }
}
es.create_index(index, mapping)
```

cron/eventStats.py
```diff
@@ -79,12 +80,16 @@ def main():
...
     logger.debug(options)
     es = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
+    index = options.index
     stats = esSearch(es)
     logger.debug(json.dumps(stats))
     try:
-        # post to elastic search servers directly without going through
-        # message queues in case there is an availability issue
-        es.save_event(body=json.dumps(stats), doc_type='mozdefstats')
+        while es.index_exists(index) == False:
```
pwnbus: I don't think we want this to run indefinitely like this, maybe we should have some sort of counter to keep track of how many times we've slept for, waiting for the index to get created? And after maybe the 3rd time, the script just exists with an error message saying the index isn't created?

tests/lib/test_elasticsearch_client.py
```diff
@@ -365,6 +365,20 @@ def test_get_indices(self):
         indices.sort()
         assert indices == [self.alert_index_name, self.previous_event_index_name, self.event_index_name, 'test_index']
 
+class TestIndexExists(ElasticsearchClientTest):
```
pwnbus: Very nicely done!

cron/eventStats.py
Phrozyn: This change has been committed.

cron/mozdefStateMappingTemplate.json
pwnbus: Understood, since the mapping should be only a couple fields, would it be better to put that dictionary inside the cron script, instead of a separate file?

Phrozyn: Right now this only has mozdefstats fields, this mapping needs changed to include the mozdefhealth fields as well.
I'll add the additional fields and we'll see just how big it will be.

Also please note the sqs queues I'm adding will also be additional mappings. Let's decide once I have the finished mapping fleshed out.

Phrozyn: I've updated the mapping, and we should probably talk about this more once you've reviewed the sqs status PR, as any changes there may indicate a need for changes here - depending on whether we are ok with the queues being mapped dynamically to a field rather than a json blob.

Phrozyn:
@pwnbus can you validate the changes and remove the request changes flag if it looks good?
This must be merged before the sqs stats PR.

cron/healthAndStatus.py
```diff
@@ -156,6 +167,11 @@ def initConfig():
     options.esservers = list(getConfig('esservers',
                                        'http://localhost:9200',
                                        options.configfile).split(','))
+    # configure the index to save events to
+    options.index = getConfig('index', 'mozdefstate', options.configfile)
+    # point to mapping json for the index
+    default_mapping_location = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'mozdefStateMappingTemplate.json')
```
Phrozyn: Found a bug here, this filename doesn't match the one committed, will push an update for this after lunch

Phrozyn: Seems I committed 2 files for mapping, going to remove one in favor of the other. 
mozdefStateMappingTemplate.json will be removed
mozdefStateDefaultMappingTemplate.json will be retained.

cron/mozdefStateDefaultMappingTemplate.json
```diff
@@ -0,0 +1,387 @@
...
+                    "long_fields": {
+                        "match": "*",
+                        "match_mapping_type": "long",
+                        "mapping": {
+                            "type": "keyword"
+                        }
+                    }
+                }
+            ],
+            "properties": {
```
pwnbus: Since most of these fields aren't used in the state index, can we remove the unnecessary ones?

cron/eventStats.py
```diff
@@ -79,12 +80,21 @@ def main():
...
     es = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
+    index = options.index
     stats = esSearch(es)
     logger.debug(json.dumps(stats))
+    sleepcycles = 0
     try:
-        # post to elastic search servers directly without going through
-        # message queues in case there is an availability issue
-        es.save_event(body=json.dumps(stats), doc_type='mozdefstats')
+        while es.index_exists(index) == False:
```
pwnbus: The preferred way of doing this type of logic is to do:
```
while not es.index_exists(index):
   ...
```

cron/eventStats.py
```diff
@@ -79,12 +80,21 @@ def main():
...
+            sleep(3)
+            if sleepcycles == 3:
+                logger.debug("The index is not created. Terminating eventStats.py cron job.")
+                exit(1)
+            sleepcycles += 1
+            if es.index_exists(index) == True:
```
pwnbus: A more pythonic way to do this is:
```
if es.index_exists(index):
    ...
```

cron/eventStats.py
Phrozyn: :) thanks! I'll correct.

cron/eventStats.py
Phrozyn: Interesting, I did not know this.

PR #728, opened 2018-07-21
Adding new sqs status script for mozdefhealth

Phrozyn:
This is pointed to the new mozdefstate index, so it should not be merged until PR #723 is merged.

This is a work in progress and I am sure  there will be requests for changes!

### DIFF
--- a/cron/sqs_status.conf
+++ b/cron/sqs_status.conf
@@ -1,7 +1,8 @@
 [options]
+account = <infosec-{dev|prod}>
 accesskey = <accesskey>
 secretkey = <secretkey>
-region = <region>
-taskexchange = sqs,comma,delimited,list
+region = sqs-region
+taskexchange = comma,delimited,sqs,queues
 esservers = http://localhost:9200
 index = mozdefstate
--- a/cron/sqs_status.conf
+++ b/cron/sqs_status.conf
@@ -3,3 +3,5 @@ accesskey = <accesskey>
 secretkey = <secretkey>
 region = <region>
 taskexchange = sqs,comma,delimited,list
+esservers = http://localhost:9200
+index = mozdefstate
--- a/cron/sqs_status.conf
+++ b/cron/sqs_status.conf
@@ -1,0 +1,5 @@
+[options]
+accesskey = <accesskey>
+secretkey = <secretkey>
+region = <region>
+taskexchange = sqs,comma,delimited,list
--- a/cron/sqs_status.conf
+++ b/cron/sqs_status.conf
@@ -1,8 +1,7 @@
 [options]
-account = <infosec-{dev|prod}>
 accesskey = <accesskey>
 secretkey = <secretkey>
-region = sqs-region
-taskexchange = comma,delimited,sqs,queues
+region = <region>
+taskexchange = sqs,comma,delimited,list
 esservers = http://localhost:9200
 index = mozdefstate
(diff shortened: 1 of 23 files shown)


cron/sqs_status.conf
```diff
@@ -0,0 +1,7 @@
+[options]
+accesskey = <accesskey>
+secretkey = <secretkey>
+region = <region>
+taskexchange = sqs,comma,delimited,list
```
pwnbus: We need the ability to specify an account owner, per region, per sqs queue. What we have right now, will not cover all of the queues that we care about (as we are restricting our "search" to only one region for one account).

jeffbryner: wouldn't it be the same logic/script, just a different -c conf file for each queue you are targeting?

cron/sqs_status.py
```diff
@@ -0,0 +1,116 @@
...
+    options.index = getConfig('index', 'mozdefstate', options.configfile)
+
+if __name__ == '__main__':
+    # configure ourselves
+    parser = OptionParser()
+    parser.add_option("-c", dest='configfile', default=sys.argv[0].replace('.py', '.conf'), help="configuration file to use")
+    (options, args) = parser.parse_args()
+    initConfig()
+    initLogger(options)
+    getQueueSizes()
```
jeffbryner: shouldn't this call main() instead of doing what main() is doing?

cron/sqs_status.conf
pwnbus: I'd argue that it'd be better to modify a config file to add a new queue, instead of having to append to crontab to run another python file.

jeffbryner: Then the same access key/secret key/region has to be valid for each queue?

cron/sqs_status.py
Phrozyn: You're absolutely right.

cron/sqs_status.conf
Phrozyn: Yeah this is a concern, I was sort of modeling this off of how we currently do healthAndStatus, but what I noticed is that even healthAndStatus only seems to pull data and insert an event from a single host at a time, that data is not merged together in the result as I was trying to do here.

SQS queues are not host based, and I think that's where I'm finding some issue with how to extract this data and present it without it causing some logic problems.

I did want to avoid a separate config for each queue, but doing this will require a better schema structure, and if everything is in one event, then the event can possibly have an infinite amount of fields in it's mapping (around 3 or 4 for each queue).

Also to speak to @jeffbryner's point of accesskey/secretkey for all of them does cause problems, because we don't have the same key in multiple accounts. 

So with that said, I think moving back to the one config per queue addresses most of these problems, and if one queue fails, it doesn't fail for all of them in this manner.

I can add ownership information for each queue as well to address that need.

I do feel there is a more elegant way to do this, but not with crons, I'm thinking more of a microservice implementation that would run on supervisord to kick off these checks and feed into ES using celery maybe. Could possibly move to that kind of structure for all of our crons and get out of using cron and simply add the queue name, keys, region, and owner to a list in a celery config, or use existing esworker configs to grab the required data and populate the values needed for the connection. 
So instead of having a cron directory for these items, maybe we just have a plugin that cycles through the mq configs and pulls this data using celery workers?

pwnbus: Would it be better to represent the config file in a json format? So that you can do something like:

```
queues = [
    {
        "queue_name" : "ExampleQueue",
        "access_key": "12345",
        "secret_key": "abcdefg",
        "region": "us-west1",
    },
    {
        "queue_name" : "AnotherExampleQueue",
        "access_key": "678910",
        "secret_key": "hijklmnop",
        "region": "us-west2",
    }
]
```

Phrozyn: Going to try the json config and see if I can make that work.

Phrozyn:
Closing this PR since it's more efficient to just open a new one from a new branch. I can't seem to locate the commit that added all the merged commits into this PR.

PR #731, opened 2018-08-06
Adding sqs_queue_status script to cron

Phrozyn:
I didn't do a json config for all accounts (we only have 2) 
In the interest of time and needing to get to goal items, doing so would require more time and refactoring as it's something I am unfamiliar with.

### DIFF
--- a/cron/sqs_queue_status.py
+++ b/cron/sqs_queue_status.py
@@ -117,7 +117,7 @@ def getQueueSizes():
     # post another doc with a static docid and tag
     # for use when querying for the latest sqs status
     healthlog['tags'] = ['mozdef', 'status', 'sqs-latest']
-    es.save_event(index=options.index, doc_id=getDocID(sqsid), body=json.dumps(healthlog))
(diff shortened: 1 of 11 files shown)


cron/sqs_queue_status.py
```diff
@@ -0,0 +1,150 @@
...
+
+def getDocID(sqsregionidentifier):
+    # create a hash to use as the ES doc id
+    # hostname plus salt as doctype.latest
+    hash = md5()
+    hash.update('{0}.mozdefhealth.latest'.format(sqsregionidentifier))
+    return hash.hexdigest()
+
+def getQueueSizes():
+    logger.debug('starting')
```
zsck: Would it be useful for this to log something like "starting sqs_queue_status getQueueSizes"?

zsck: I don't think we should continue using an arguably suboptimal style  just because it's standardized. Adding more information here will only serve as an improvement and a reminder that we should make similar changes elsewhere.  My experience, and it seems our team's experience, with delaying such changes for a future concerted effort is that this future effort is never made.

PR #736, opened 2018-08-17
Modify unknown isp in geomodel plugin

pwnbus:


alerts/plugins/dashboard_geomodel.py
```diff
@@ -82,13 +82,14 @@ def onMessage(self, message):
...
 
-        source_ip_whois = ""
+        new_ip_info = ""
         try:
             whois = IPWhois(source_ip).lookup_whois()
             whois_str = whois['nets'][0]['description']
-            source_ip_whois = whois_str.replace('\n', ' ').replace('\r', '')
+            source_ip_isp = whois_str.replace('\n', ', ').replace('\r', '')
+            new_ip_info = u'{} ({})'.format(source_ip, source_ip_isp)
         except Exception:
```
andrewkrug: What specific exception are you trying to catch here?  This will bury literally everything.   

Generally I like to use 

```
except Exception as e:
   logger.error({}.format(e)) 
```

As a minimum so you can know why.

PR #746, opened 2018-09-13
hostname field normalization phase I 

Phrozyn:
details.hostname is a field added due to heka parsing. This is to normalize hostname data fields after decom of heka within our pipeline. The top level mandatory field will be the only field to contain the event source hostname for eventtask events that are parsed. Other hostname fields will be reviewed in a later phase in other data sources.

Many MQ Plugins add the hostname to the details.hostname field. There are also alerts which look for details.hostname to obtain the hostname for an alert. I propose that we add code to look for or validate the top level hostname field, and leave the details.hostname field in the plugin if applicable until we validate that absolutely nothing is using it. Once we can validate that nothing is coming in with details.hostname populated, we can remove that portion of the codefrom our plugins. All alerts that use the details.hostname field do have the top level hostname field and they match. So no need to keep the details.hostname field around in those particular scripts.

The only real modification where we look for both is in the mozilla_location plugin.
This would simply be an addition of an elif, or, and if statements looking for hostname in the top level in addition to the one looking for it in details.hostname. Any plugins that define details.hostname will be changed to just be hostname in the top level.

### DIFF
--- a/alerts/auditd_sftp.py
+++ b/alerts/auditd_sftp.py
@@ -34,9 +34,9 @@ class AlertSFTPEvent(AlertTask):
         username = 'unknown'
         directory = 'unknown'
         x = event['_source']
+        if 'hostname' in x:
+            srchost = x['hostname']
         if 'details' in x:
-            if 'hostname' in x['details']:
-                srchost = x['details']['hostname']
             if 'originaluser' in x['details']:
                 username = x['details']['originaluser']
             if 'cwd' in x['details']:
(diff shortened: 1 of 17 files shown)


Phrozyn:
Initial commit is for mq plugins that define details.hostname

Phrozyn:
Validated that all events that trigger these alerts have hostname populated, while not all of them have details.hostname populated

Phrozyn:
No further edits needed.

PR #747, opened 2018-09-18
Add alert for honeycomb

caggle:
First attempt in creating a MozDef alert.

### DIFF
--- a/alerts/honeycomb.py
+++ b/alerts/honeycomb.py
@@ -17,7 +17,8 @@ class AlertHoneycomb(AlertTask):
 
         search_query.add_must([
             TermMatch('category', 'syslog'),
-            TermMatch('processname', 'Honeycomb')
+            TermMatch('processname', 'Honeycomb'),
+            TermMatch('severity', 'CRIT')
         ])
 
         self.filtersManual(search_query)
--- a/alerts/honeycomb.py
+++ b/alerts/honeycomb.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+from lib.alerttask import AlertTask
+from query_models import SearchQuery, TermMatch
+import re
+
+
+class AlertHoneycomb(AlertTask):
+    def main(self):
+        # look for events in last 10 mins
+        search_query = SearchQuery(minutes=10)
+
+        search_query.add_must([
+            TermMatch('category', 'syslog'),
+            TermMatch('processname', 'Honeycomb')
+        ])
+
+        self.filtersManual(search_query)
+
+        # Search aggregations on field 'hostname', keep X samples of
+        # events at most
+        self.searchEventsAggregated('hostname', samplesLimit=10)
+        # alert when >= X matching events in an aggregation
+        # I think it makes sense to alert every time here
+        self.walkAggregations(threshold=1)
+
+    # Set alert properties
+    def onAggregation(self, aggreg):
+        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
+        # aggreg['value']: value of the aggregation field, ex: toto@example.com
+        # aggreg['events']: list of events in the aggregation
+        category = 'honeypot'
+        tags = ['honeypot', 'honeycomb']
+        severity = 'WARNING'
+
+        # This is required in order to extract and display the source IP accessing the honeypot
+        pattern = '/originating_ip="(\d+\.\d+\.\d+\.\d+)"/'
+        offendingIPs = []
+
+        for event in aggreg['allevents']:
+            offendingIPs.append(re.search(pattern, aggreg['allevents'][event]['_source']['summary']))
+
+        summary = 'Honeypot activity on {0} from IP(s): {1}'.format(aggreg['value'], ",".join(set(offendingIPs)))
+
+        # Create the alert object based on these properties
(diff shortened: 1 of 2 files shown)


alerts/honeycomb.py
```diff
@@ -0,0 +1,51 @@
...
+        category = 'honeypot'
+        tags = ['honeypot', 'honeycomb']
+        severity = 'WARNING'
+
+        # This is required in order to extract and display the source IP accessing the honeypot
+        pattern = '/originating_ip="(\d+\.\d+\.\d+\.\d+)"/'
+        offendingIPs = []
+
+        for event in aggreg['allevents']:
+            offendingIPs.append(re.search(pattern, aggreg['allevents'][event]['_source']['summary']))
```
claudijd: I wonder if we need some more logic here to cover a case where the data within summary is not what we expect.

Say summary contained "foo", would this stack trace mozdef or handle it gracefully?

alerts/honeycomb.py
```diff
@@ -0,0 +1,51 @@
...
+
+        summary = 'Honeypot activity on {0} from IP(s): {1}'.format(aggreg['value'], ",".join(set(offendingIPs)))
```
claudijd: @pwnbus on alert logic that might produce "N" sources of malicious activity, do we need to worry about producing a summary below a specific length to avoid run-off?

tests/alerts/test_honeycomb_case.py
```diff
@@ -0,0 +1,74 @@
...
+    alert_filename = "honeycomb"
+
+    # This event is the default positive event that will cause the
+    # alert to trigger
+    default_event = {
+        "_type": "event",
+        "_source": {
+            "category": "syslog",
+            "details": {
+                "processname": "Honeycomb"
```
claudijd: My guess is that this default event needs summary data if the alert and aggregation is consuming this information.

We should probably also consider providing two alternative examples with one from say IP A and another from IP B to ensure that our aggregated source IPs are properly being served.

Additionally, as we noted in testing, there's a weird quote chars being presented in the summary from honeycomb and we should carry these forward in our test cases.

alerts/honeycomb.py
caggle: I think you mean, if the event summary is completely different than what expect, i.e. no originatingIP, but simply a random string like "foo". Still I don't think it would stack trace (not from regex at least), but we should probably but a validation the next line in case there is no match (i.e. len(offendingIP) == 0)

caggle: I have included a condition from L48.

alerts/honeycomb.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class AlertHoneycomb(AlertTask):
+    def main(self):
+        # look for events in last 10 mins
+        search_query = SearchQuery(minutes=10)
+
+        search_query.add_must([
+            TermMatch('category', 'syslog'),
+            TermMatch('processname', 'Honeycomb')
```
claudijd: Can we add a requirement that the logging priority be critical?  Honeycomb does emit other severity levels, but criticals are the only ones we're really interested in for the sake of alerting.

alerts/honeycomb.py
claudijd: This is actually the condition I was thinking about (though not a stack trace, certainly non-desirable)...

>>> array = [1,2,3]
>>> array.append(re.search('foo', 'bar'))
>>> array
[1, 2, 3, None]

caggle: Added another condition in https://github.com/mozilla/MozDef/pull/747/commits/81738efeffecad9d288589a63905f37e4285e87e

alerts/honeycomb.py
caggle: Done in https://github.com/mozilla/MozDef/pull/747/commits/aff07d46fcd5496e0d36676b111ef3bd146757a5

alerts/honeycomb.py
```diff
@@ -0,0 +1,58 @@
...
+        severity = 'WARNING'
+
+        # This is required in order to extract and display the source IP accessing the honeypot
+        pattern = '/originating_ip="(\d+\.\d+\.\d+\.\d+)"/'
+        offendingIPs = []
+
+        for event in aggreg['allevents']:
+            sourceIP = re.search(pattern, aggreg['allevents'][event]['_source']['summary'])
+            if sourceIP is None:
+                pass
```
claudijd: I think 'continue' is more appropriate in this context

https://www.tutorialspoint.com/python/python_loop_control.htm

claudijd:
I made a couple adjustments to this branch to bring add essential fields to the example event defaults as well as adding some test cases that stress the conditionals in the aggregation logic of the alert.  It's still not passing, but I think it's really close.

alerts/honeycomb.py
```diff
@@ -0,0 +1,56 @@
...
+        self.walkAggregations(threshold=1)
+
+    # Set alert properties
+    def onAggregation(self, aggreg):
+        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
+        # aggreg['value']: value of the aggregation field, ex: toto@example.com
+        # aggreg['events']: list of events in the aggregation
+        category = 'honeypot'
+        tags = ['honeypot', 'honeycomb']
+        severity = 'WARNING'
```
pwnbus: Just noticed this, but do we want to make this severity CRITICAL?

caggle: From memory, making it critical is the eventual goal but we'd like to get it more "productionized" first by possibly installing more instances/more services.

claudijd: Yeah, I think warning for a few weeks and triggering examples of it and seeing it live will help increase confidence before making critical

pwnbus: Works for me

PR #748, opened 2018-09-18
Add proxy drop on executable

claudijd:


alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,53 @@
...
+
+class AlertProxyDropExecutable(AlertTask):
+    def main(self):
+        search_query = SearchQuery(minutes=5)
+
+        search_query.add_must([
+            TermMatch('category', 'squid'),
+            TermMatch('details.proxyaction', 'TCP_DENIED/-'),
+            ExistsMatch('details.proxyaction'),
+            QueryStringMatch('details.destination: /\.(exe|bin|sh|py|rb)$/')
```
claudijd: I'm pretty sure this regex doesn't work, needs to be verified as proper syntax

cknowles-moz:
With increased looking at TCP_DENIED, is there a way to setup some sort of "exception list" - thinking here of the known bug around centos mirrors popping into existence... it's a known issue that's being worked on, but it causes a significant amount of TCP_DENIED noise in the proxies.

claudijd:
@Phrozyn yes, totally (RE: unit-tests), just wanted to get the concept up for consideration.  If all seems good and we have it where we want it, then I'll add the tests to make it so.

alerts/proxy_drop_executable.conf
```diff
@@ -0,0 +1,2 @@
+[options]
+destinatationfilter = (repomd.xml$)
```
claudijd: This is to work around an issue where by CentOS will ignore provided repositories.  We will not alert on those cases to try and get a better signal to noise.  There is not much concern around an attacker knowing this, as the traffic is being dropped anyways.

claudijd:
I think I want to change the name of this alert, because I have a few other ideas for suspicious proxy drop behavior that I want to capture.  I think repeated drops is an interesting thing, especially when we add the executable context, but in addition to that I think it would be interesting to have yet another alert that suggests CONNECT proxy attempt abuse.

For example:

TCP_DENIED's can come from HTTP/HTTPS traffic, but their destinations will have slightly different attributes (80's will have URLs, 443's will have host/port), what if we had an additional alert that triggered on denied events for non-std ports. 

TCPDENIED for say example.com:444

I recall an old trick I used to use to get IRC bots to push through a connect proxy to call home and although we would obviously block that, it would be nice to have a more upfront alert suggesting this is deny behavior we should be looking closer at to detect exfil/cnc behavior.

Phrozyn:
I kind of like the idea of two separate alerts, but the proxy doesn't support non-std ports (outside of 80/443) at all. This could be quite noisy as individuals set up systems that are configured to use the proxy might have things that need to have flows opened and the proxy bypassed. However, it can also alert us to nefarious programs trying to phone home. It's a double-edged sword. I definitely think this would be one of those alerts MOC should handle as a first response.

claudijd:
@Phrozyn the proxy does support non-std ports by function, just not by policy.  So these will be drops in #750 and I included an example to demonstrate this.

In this particular PR, the query logic restricts to events with executable extensions that are already being denied, this is bubbling up behavior to make us aware of that fact, which would otherwise go unnoticed in the regular deny noise, which makes up nearly all traffic.

alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,63 @@
...
+            TermMatch('tags', 'squid'),
+            TermMatch('details.proxyaction', 'TCP_DENIED/-'),
+            QueryStringMatch(
+                'details.destination: /{}/').format('.*\.(exe|bin|sh|py|rb)$')
+        ])
+
+        # search_query.add_must_not([
+        #     QueryStringMatch(
+        #         'details.destination: /{}/'.format(self.config.destinationfilter)),
+        # ])
```
claudijd: This is being commented out for the time being for TS purposes.

alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,63 @@
...
+    # Set alert properties
+    def onAggregation(self, aggreg):
+        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
+        # aggreg['value']: value of the aggregation field, ex: toto@example.com
+        # aggreg['events']: list of events in the aggregation
+        category = 'squid'
+        tags = ['squid', 'proxy']
+        severity = 'WARNING'
+
+        summary = "Multiple Proxy DROP events detected from 1.2.3.4 to the following executable file destinations: http://evil.com/evil.exe"
```
claudijd: This is a temporary short circuit this until main is fixed.

alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,83 @@
...
+        # for extension in self.config.extensions.split(','):
+        #     if run == 0:
+        #         superquery = QueryStringMatch(
+        #             'details.destination: /{0}/'.format(extension))
+        #     else:
+        #         superquery |= QueryStringMatch(
+        #             'details.destination: /{0}/'.format(extension))
+        #     run += 1
+        #
+        # search_query.add_must(superquery)
```
claudijd: @pwnbus please note that there is something in this logic which does not cause an exception but prevents a match.  Initially, I thought it was somehow AND'ing instead of OR'ing, but that does not appear to be the case based on a test where I left only "exe" in the config exceptions.  Something else is happening here.

alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,83 @@
...
+        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
+        # aggreg['value']: value of the aggregation field, ex: toto@example.com
+        # aggreg['events']: list of events in the aggregation
+        category = 'squid'
+        tags = ['squid', 'proxy']
+        severity = 'WARNING'
+
+        dropped_url_destinations = []
+
+        # START HACK - The conditionals here are to account for an inability to endwith
```
claudijd: This hack is implemented to show we can make the unit-tests pass without fixing the query search.  The concern however is that by not fixing this we can run into a situation where a given source IP has multiple deny events within an aggregated search and the ones we want are discarded (or not sampled) and then we further weed those events in the onAggregation such that a exe target failure would basically not fire in some circumstances.  Not to mention, it's horribly unperformant to do this in python and not in Lucene/ES (at least in concept, I'm not sure how much more overhead this presents).

alerts/proxy_drop_executable.py
```diff
@@ -0,0 +1,52 @@
...
+        ])
+
+        self.filtersManual(search_query)
+
+       # Search aggregations on field 'hostname', keep X samples of
+        # events at most
+        self.searchEventsAggregated('details.sourceipaddress', samplesLimit=10)
+        # alert when >= X matching events in an aggregation
+        # I think it makes sense to alert every time here
+        self.walkAggregations(threshold=1)
```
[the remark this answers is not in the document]
claudijd: Discussed with pwnbus, this is fine.

PR #750, opened 2018-09-24
Add connect proxy drops to non-std ports alert

claudijd:
This aims to capture occurances where an attacker is attempting (and failing) to egress via the proxy and making it an alert that IR responders can act on.

### DIFF
--- a/alerts/proxy_drop_non_standard_port.conf
+++ b/alerts/proxy_drop_non_standard_port.conf
@@ -1,0 +1,2 @@
+[options]
+excludedports = 443,8443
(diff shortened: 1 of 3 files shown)


Phrozyn:
Just a note as I stated in PR 748 that this can result in high false positives due to the nature of the proxy. Maybe to start we put this in INFO level and then raise if we find that it's not as FP heavy as I suspect it might be.

claudijd:
I don't believe high-false positives are a concern here.  Namely because we have a not-statement on the query syntax on "443", which is the only port that should legitimately using the proxy.  Additionally, we have a context to restrict this to denies, such that it's really focused on catching malicious behavior more so than misconfig behavior.  This is actually a common tactic with IRC bots phone-home behavior.

I also verified that non-std ports on the connect proxy are loggged:

screen_shot_2018-09-25_at_4_02_10_pm

claudijd:
@pwnbus I want to hold on this, I have some tweaks add more port exceptions, or we should move to a config

PR #754, opened 2018-10-01
Fixing details.dhost to be hostname

Phrozyn:
in these events, hostname is normally populated as UNKNOWN, while details.dhost is populated with the hostname. This PR corrects this by moving details.dhost into the hostname field.

### DIFF
--- a/mq/plugins/auditdFixup.py
+++ b/mq/plugins/auditdFixup.py
@@ -102,7 +102,7 @@ class message(object):
                 del message['details']['gid']
 
        # fix details.dhost to be hostname
-       if 'details' in message.key() and isinstance(message['details'], dict):
+       if 'details' in message.keys() and isinstance(message['details'], dict):
            if 'dhost' in message['details'].keys():
                # details.dhost is the host that the auditd event is happening on.
                message['hostname'] = message['details']['dhost']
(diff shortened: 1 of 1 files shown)


mq/plugins/auditdFixup.py
```diff
@@ -101,6 +101,13 @@ def onMessage(self, message, metadata):
                 message['details']['gidstring'] = message['details']['gid']
                 del message['details']['gid']
 
+       # fix details.dhost to be hostname
+       if 'details' in message.key() and isinstance(message['details'], dict):
```
pwnbus: This line has a syntax error, there is no method called "key()"
What you want instead, is to do something like this:
```
if 'details' in message:
```

The same goes for when you verify if 'dhost' is in message['details']

Phrozyn: it's a typo, should have been message.keys() which is in other places in this script, with that said, does that mean that those other fixups that contain message.keys() are not working? I've used message.keys() in other places and it does work in that regard.

Phrozyn: I just checked other events where message.keys() is used to transform a field, and the other locations that this is present does appear to be functional, so I think just correcting the typo is all that's needed here.

pwnbus: This if statement works just fine:
```
if 'details' in message.keys():
```
The proposed syntax works the same as above, just seems to be more pythonic. With that being said, the typo fix will be sufficient

PR #755, opened 2018-10-02
If cluster_client_ip is present there seems to be no reason to use th…

mpurzynski:
…e sourceipaddress. The cluster_client_ip should overwrite as the 'true' client's IP. This is to enable anomaly detection, like Geo, on traffic going through load balancers.


Phrozyn:
I agree with @pwnbus' comment above, I think we should dig into this a little further and determine whether further normalization is needed for the data we are trying to overwrite here.

mpurzynski:
If the sourceipaddress field is present together with the cluster_client_ip field then the later should be used and the former overwritten. Please note this is very different than the x-forwarded-for header, that can easily be spoofed.

PR #769, opened 2018-10-08
docker nginx proxy pass fix

Phrozyn:
nginx, by default, only resolves a proxy_pass one time unless it contains a variable. That means that if meteor, loginput, restapi or kibana are restarted that nginx would have to be restarted as well in order to get the new internal container ip address of the upstream.  Moving all of the proxy_pass upstreams into a variable forces nginx to resolve them.

### DIFF
--- a/docker/compose/nginx/files/nginx.conf
+++ b/docker/compose/nginx/files/nginx.conf
@@ -9,6 +9,7 @@ events {
 http {
   include /etc/nginx/mime.types;
   default_type application/octet-stream;
+  resolver 127.0.0.11;
 
   log_format main
           '$remote_addr - $remote_user [$time_local] '
--- a/docker/compose/nginx/files/nginx.conf
+++ b/docker/compose/nginx/files/nginx.conf
@@ -51,8 +51,10 @@ http {
     listen              *:80;
     access_log          off;
 
+    set $meteor_proxy_pass_url http://meteor:3000;
+
     location /{
-      proxy_pass http://meteor:3000;
+      proxy_pass $meteor_proxy_pass_url;
       proxy_read_timeout 90;
     }
   }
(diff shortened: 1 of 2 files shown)


docker/conf/nginx.conf
```diff
@@ -48,8 +48,11 @@ http {
   server{
     listen              *:80;
     access_log          off;
+
+		set $meteor_proxy_pass_url http://127.0.0.1:3000;
```
pwnbus: Indentation is off

docker/conf/nginx.conf
```diff
@@ -60,9 +63,11 @@ http {
     server_name         localhost;
     access_log          off;
 
+		set $kibana_proxy_pass_url http://127.0.0.1:5601;
```
pwnbus: Same here, I think tabs vs spaces issue perhaps?

Phrozyn: Maybe, I'll see if I can figure out what's going on between my ide and me, will correct shortly.

Phrozyn: I know what is causing this. They are set to spaces in the IDE, and are indented correctly there, but because I submitted the commit from my windows machine git is trying to correct for crlf I think that may be throwing things off.

Phrozyn: I just pulled it down on the mac, and indentation is correct there too for these files.
Whatever github is showing is incorrect.

docker/conf/nginx.conf
Phrozyn: looks fine in vscode, not sure what github is doing to my commit :( 

Going to correct online rather than in my branch since it's showing correct there.

docker/compose/nginx/files/nginx.conf
```diff
@@ -9,6 +9,7 @@ events {
 http {
   include /etc/nginx/mime.types;
   default_type application/octet-stream;
+  resolver 127.0.0.11;
```
pwnbus: This won't be consistent every time, the only reason we have resolver specified in prod is for auth0 integration, so I don't think we need this line.

Phrozyn: This is a docker embedded construct for  the nginx to be able to query DNS for the other containers, it's always this ip.  This has no relation to the resolver we have in the production nginx config. We add this because we have our networks defined.

https://docs.docker.com/v17.09/engine/userguide/networking/configure-dns/

PR #770, opened 2018-10-08
Meteor 1.7 update

jeffbryner:
This updates the meteor UI to be compatible with the latest meteor 1.7.0.3, incorporating: 

- Directory/code placement standards
- ES2015 import declarations
- Updating libraries used by MozDef


docker/compose/docker-compose-cron.yml
```diff
@@ -0,0 +1,22 @@
...
+version: '2.2'
+services:
+  cron:
+    build:
+      context: ../../
+      dockerfile: docker/compose/mozdef_cron/Dockerfile
+    restart: always
+    command: bash -c 'sleep 90 && crond -n'
+    volumes:
+      - /Users/jbryner/development/mozdef/deployments/mozillagit/cron:/opt/mozdef/envs/mozdef/cron
```
pwnbus: Hard-coded paths here

docker/compose/docker-compose-meteor-ui.yml
```diff
@@ -0,0 +1,18 @@
...
+      dockerfile: docker/compose/mozdef_meteor/Dockerfile
+    restart: 'no'
+    command: bash -c 'node -i'    
+    depends_on:
+      - mongodb
+      - rest
+    networks:
+      - default
+    volumes:
+      - /Users/jbryner/development/mozdef/deployments/mozillagit/meteor:/opt/mozdef/envs/mozdef/meteor
```
pwnbus: Same here, hardcoded paths

docker/compose/docker-compose-rest.yml
```diff
@@ -0,0 +1,20 @@
...
+    command: bash -c 'source /opt/mozdef/envs/python/bin/activate && python index.py -c index.conf'
+    depends_on:
+      - base
+      - elasticsearch
+      - mongodb
+      - bootstrap
+    networks:
+      - default
+    volumes:
+      - /Users/jbryner/development/mozdef/deployments/mozillagit/rest:/opt/mozdef/envs/mozdef/rest
```
pwnbus: And here.

docker/compose/live-reload.override.yml
```diff
@@ -0,0 +1,36 @@
+---
+version: '2.2'
+services:
+  meteor:
+    build:
+      context: ../../
+      dockerfile: docker/compose/mozdef_meteor/Dockerfile
+    restart: always
+    # command: bash -c 'cd /opt/mozdef/envs/mozdef/meteor && meteor npm install --save-dev && rm -rf .meteor/local && /usr/bin/meteor'
```
pwnbus: Can we delete the commented out line here or is there some value in keeping it around? I've got commented out lines in other docker-compose files to make it easy to expose ports for example, so if it's not leftover cruft, we can keep it around.

docker/compose/live-reload.override.yml
```diff
@@ -0,0 +1,36 @@
...
+    command: bash -c 'node -i'
+    depends_on:
+      - mongodb
+      - rest
+    networks:
+      - default
+    volumes:
+        - ../../meteor:/opt/mozdef/envs/mozdef/meteor
+#        - ../../meteor/private:/opt/mozdef/envs/mozdef/meteor/private
```
pwnbus: Are these 2 commented lines still needed?

docker/compose/mozdef_base/Dockerfile
```diff
@@ -17,14 +17,9 @@ RUN \
...
   mkdir /opt/mozdef/envs && \
-  curl -s -L https://www.python.org/ftp/python/$PYTHON_VERSION/Python-$PYTHON_VERSION.tgz  | tar -C /opt/mozdef/ -xz && \
-  cd /opt/mozdef/Python-$PYTHON_VERSION && \
-  ./configure && \
-  make && \
-  make install && \
-  rm -r /opt/mozdef/Python-$PYTHON_VERSION && \
   cd /opt/mozdef && \
   yum install -y mysql-devel \
+                    python \
```
pwnbus: Is there any worry here that centos7 will update to python3, and thus we'd run into breaking changes? Can we make sure we install python27 via yum?

Also, we can delete the $PYTHON_VERSION environment variable at the top of this file.

docker/compose/mozdef_meteor/Dockerfile
```diff
@@ -22,10 +22,15 @@ RUN \
                 gcc-c++ \
                 libstdc++ \
                 libffi-devel \
-                zlib-devel && \
-  curl -sL -o /opt/mozdef/nodesource.rpm https://rpm.nodesource.com/pub_4.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
-  rpm -i --nosignature --force /opt/mozdef/nodesource.rpm && \
-  yum install -y nodejs-$NODE_VERSION && \
+                zlib-devel
+# RUN \
+#   curl -sL -o /opt/mozdef/nodesource.rpm https://rpm.nodesource.com/pub_4.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
```
pwnbus: Can we delete these commented out lines?

docker/compose/mozdef_meteor/Dockerfile
```diff
@@ -39,12 +44,12 @@ RUN chown -R mozdef:mozdef /opt/mozdef/envs/mozdef/meteor
 
 USER mozdef
 RUN \
-  mkdir -p /opt/mozdef/envs/meteor/mozdef && \
-  cd /opt/mozdef/envs/mozdef/meteor && \
-  meteor npm install && \
-  meteor build --server localhost:3002 --directory /opt/mozdef/envs/meteor/mozdef && \
-  mv /opt/mozdef/envs/mozdef/meteor/node_modules /opt/mozdef/envs/meteor/mozdef/node_modules
+  mkdir -p /opt/mozdef/envs/meteor/mozdef
+  #cd /opt/mozdef/envs/mozdef/meteor && \
```
pwnbus: Same here, can we nuke these commented out lines?

docker/compose/mozdef_sampledata/files/supervisor.conf
```diff
@@ -0,0 +1,9 @@
+[supervisord]
+nodaemon=true
+autostart=true
+autorestart=true
+#real time constant stream of demo events/alerts/attackers
+[program:realTimeEvents]
+priority=10
+command=bash -c 'sleep 90 && source /opt/mozdef/envs/python/bin/activate && /opt/mozdef/envs/mozdef/examples/demo/sampleevents.sh'
```
pwnbus: Do we need to use supervisor here since we're only running one command? Can this be done all via the command statement in docker-compose?

meteor/.meteor/packages
```diff
@@ -1,32 +1,33 @@
 # Meteor packages used by this project, one per line.
+# Check this file (and the other files in this directory) into your repository.
```
pwnbus: What's the difference between this file and meteor/.meteor/versions? For example, we have 2 versions specified for blaze-html-templates.

docker/compose/mozdef_base/Dockerfile
jeffbryner: I'm not sure of a good way to require 2.7 via yum?

docker/compose/mozdef_meteor/Dockerfile
jeffbryner: Since those build the meteor package to run with node, I added an option to the live reload overlay to trigger whether or not that step happens (will happen by default). 
https://github.com/mozilla/MozDef/pull/770/commits/efc8d2f95f0e0869f4f360ff42b8652c5abcce29

docker/compose/mozdef_sampledata/files/supervisor.conf
jeffbryner: Probably could be? Maybe there are benefits of leaving it to not require docker in order to run, though?

meteor/.meteor/packages
jeffbryner: great question, I *think* the packages file is the declarative set of intended packages where the versions is the resultant set of everything pulled in including dependencies. Dunno about the version mismatch; that's interesting but doesn't seem to affect functionality.

meteor/server/main.js
```diff
@@ -0,0 +1,350 @@
...
+import '/imports/collections.js'
+import './methods.js';
+import './mozdef.js';
+
+Meteor.startup(() => {
+  // code to run on server at startup
+
+
+if (Meteor.isServer) {
+    //Publishing setups
```
Phrozyn: Everything from line 13 to line 255 can be removed. 
These are published via collections.js in the imports directory. 
I've tested the removal in my own environment, and then tested that all the functionality still exists.
This will get rid of the "ignoring publish.. " error messages when starting meteor.

jeffbryner: good catch! missed the 'cut' part of copy/paste ;-]

docker/compose/mozdef_base/Dockerfile
andrewkrug: CentOS packaging syntax isn't going to bump python to python2.7 unless CentOS 8.  This would violate the rules of RedHat.  I think we should leave this as is.

PR #778, opened 2018-10-11
Proxy Drop Alert on IP Destinations

claudijd:
This alert would fire in cases where we have an attacker who's attempting to get out through the proxy on an IP specific destination, which is presumably suspicious.


alerts/proxy_drop_ip.py
```diff
@@ -0,0 +1,70 @@
...
+
+        search_query.add_must([
+            TermMatch('category', 'squid'),
+            TermMatch('tags', 'squid'),
+            TermMatch('details.proxyaction', 'TCP_DENIED/-')
+        ])
+
+        # Match on 1.1.1.1, http://1.1.1.1, or https://1.1.1.1
+        # This will over-match on short 3-char domains like foo.bar.baz.com, but will get weeded out below
+        ip_regex = '/.*\..{1,3}\..{1,3}\..{1,3}(:.*|\/.*)/'
```
pwnbus: I think this needs to be a bit more strict, I believe this false positives on the following details.destionation:
```
http://mirrors.cat.pdx.edu/centos/7.3.1611/os/x86_64/repodata/repomd.xml
```

claudijd: Did you note the comment on line 25?  I pretty much say this much and validate later to prevent false positives.

alerts/proxy_drop_ip.py
```diff
@@ -0,0 +1,70 @@
...
+        self.searchEventsAggregated('details.sourceipaddress', samplesLimit=10)
+        # alert when >= X matching events in an aggregation
+        # I think it makes sense to alert every time here
+        self.walkAggregations(threshold=1)
+
+    # Set alert properties
+    def onAggregation(self, aggreg):
+        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
+        # aggreg['value']: value of the aggregation field, ex: toto@example.com
+        # aggreg['events']: list of events in the aggregation
```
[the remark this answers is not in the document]
claudijd: ok, will do

alerts/proxy_drop_ip.py
pwnbus: Ahhh, no I did not.

claudijd:
I have reviewed 7 days of history on this, and this doesn't seem like a false positive concern.

PR #854, opened 2018-10-22
Infosec workweek

andrewkrug:
ConfigLib now supports environment variables.  ( This has resulted in a few failing tests related to several alerts. )
Make targets simplified and Makefile cleanup 
Replicate a common configuration pattern through the project.  Eliminate instances of hard coded ES_URL and AMQP urls.  
Everything in github.com/mozdef/lib becomes a python module ‘mozdef-utils’
Embrace docker-compose overrides rather than continue to duplicate code
Store containers for the cloud deployment in docker hub
Upgrade meteor to version 1.8  (dropped week of 10/15/18)


docker/compose/mozdef_base/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN \
                   libstdc++ \
                   libffi-devel \
                   zlib-devel \
+                  git \
```
pwnbus: Why do we need git installed?

docker/compose/mozdef_base/Dockerfile
```diff
@@ -22,10 +23,10 @@ RUN \
                     python \
                     python-devel \
                     python-pip && \
-  chown -R mozdef:mozdef /opt/mozdef/ && \
   pip install virtualenv && \
   mkdir /opt/mozdef/envs/mozdef && \
-  mkdir /opt/mozdef/envs/mozdef/cron
+  mkdir /opt/mozdef/envs/mozdef/cron && \
+  yum install -y git
```
pwnbus: We already have git installed in the beginning of the file

docker/compose/mozdef_base/Dockerfile
```diff
@@ -22,10 +23,10 @@ RUN \
                     python \
                     python-devel \
                     python-pip && \
-  chown -R mozdef:mozdef /opt/mozdef/ && \
   pip install virtualenv && \
   mkdir /opt/mozdef/envs/mozdef && \
-  mkdir /opt/mozdef/envs/mozdef/cron
+  mkdir /opt/mozdef/envs/mozdef/cron && \
```
pwnbus: We shouldn't need to do this, the cron container should be responsible for creating that directory I would think.

docker/compose/mozdef_meteor/Dockerfile
```diff
@@ -29,23 +29,23 @@ RUN \
...
     yum -y install nodejs
-RUN \
-  mkdir /opt/mozdef/meteor && \
-  curl -sL -o /opt/mozdef/meteor.tar.gz https://static-meteor.netdna-ssl.com/packages-bootstrap/$METEOR_VERSION/meteor-bootstrap-os.linux.x86_64.tar.gz && \
-  tar -xzf /opt/mozdef/meteor.tar.gz -C /opt/mozdef/meteor && \
-  mv /opt/mozdef/meteor/.meteor /opt/mozdef && \
-  rm -r /opt/mozdef/meteor && \
-  cp /opt/mozdef/.meteor/packages/meteor-tool/*/mt-os.linux.x86_64/scripts/admin/launch-meteor /usr/bin/meteor
+
+RUN mkdir /opt/mozdef/meteor
```
pwnbus: I thought we want to make these lines all under one RUN command, so that just one RUN command caches all these results (the filesize of the container is smaller as a result).

mozdef_util/docs/installation.rst
```diff
@@ -0,0 +1,51 @@
...
+From sources
+------------
+
+The sources for MozDef Util can be downloaded from the `Github repo`_.
+
+You can either clone the public repository:
+
+.. code-block:: console
+
+    $ git clone git://github.com/jeffbryner/mozdef_util
```
pwnbus: Can we point this to the mozdef repo instead of the fork?

mozdef_util/docs/installation.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Or download the `tarball`_:
+
+.. code-block:: console
+
+    $ curl  -OL https://github.com/jeffbryner/mozdef_util/tarball/master
```
pwnbus: Same here, point to the main mozilla/mozdef repo

mq/esworker_sqs.py
```diff
@@ -157,7 +156,7 @@ def esConnect():
 
 class taskConsumer(object):
 
-    def __init__(self, mqConnection, taskQueue, esConnection):
+    def __init__(self, mqConnection, taskQueue, esConnection, options):
```
pwnbus: This will throw a syntax error, we don't need the options parameter here.

requirements.txt
```diff
@@ -57,5 +57,8 @@ supervisor==3.3.1
 tzlocal==1.4
 uritemplate==0.6
 urllib3==1.22
-uwsgi==2.0.12
+uwsgi==2.0.17.1
 virtualenv==1.11.4
+moto==1.3.6
```
pwnbus: We don't actually require moto anywhere, do we need that line?

requirements.txt
```diff
@@ -57,5 +57,8 @@ supervisor==3.3.1
...
+WebTest==2.0.32
+freezegun==0.3.11
```
pwnbus: Webtest and freezegun are only needed for development purposes (running tests), since this is still specified in tests/requirements_tests.txt, can we remove it from this?

requirements.txt
andrewkrug: We will need it.  We're going to wrap the SQS tests and CloudTrail tests in moto mocks.

requirements.txt
andrewkrug: Looks like these were added by @zsck probably good to remove them here.

mq/esworker_sqs.py
andrewkrug: Are these workers not covered in the test suite?

docker/compose/mozdef_meteor/Dockerfile
andrewkrug: #wontfix

docker/compose/mozdef_base/Dockerfile
andrewkrug: I'm going to defer addressing this one.

PR #868, opened 2018-10-24
Add mq_cloudtrail docker service

gene1wood:
This should use the new environment file created in #866


pwnbus:
@gene1wood This looks good, no, you shouldn't need to add this to the rebuild docker-compose, because we only want it to run in cloudy mozdef environment.

PR #875, opened 2018-10-24
Rename generic mq_eventtask container to worker

pwnbus:


Makefile
```diff
@@ -6,7 +6,7 @@
 
 ROOT_DIR	:= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 DKR_IMAGES	:= mozdef_alertplugins mozdef_alerts mozdef_base mozdef_bootstrap mozdef_meteor mozdef_rest \
-		   mozdef_mq_eventtask mozdef_loginput mozdef_cron mozdef_elasticsearch mozdef_mongodb \
+		   mozdef_mq_worker mozdef_loginput mozdef_cron mozdef_elasticsearch mozdef_mongodb \
```
pwnbus: @gdestuynder I couldn't find a make command to build and push to docker hub, so I'm curious if I'm doing this right.

Basically, the TL;DR is that we have a container named mozdef_mq_worker, that is a generic mq worker container, that we can use to create services such as "mozdef_mq_eventtask", and mozdef_mq_cloudtrail" which uses the mozdef_mq_worker container.

gdestuynder: that seems right

PR #877, opened 2018-10-25
Remove unused collections file

pwnbus:
I believe this file is now stored in meteor/imports/collections.js


Phrozyn:
This is odd that this was in /lib, it was always in meteor under meteor/lib/
Good catch!

PR #893, opened 2018-10-27
Remove details.apiversion to mq plugin

pwnbus:
So unfortunately this didn't work as we expected, because we already specify details.apiversion as a keyword in the default mapping -> https://github.com/mozilla/MozDef/blob/master/cron/defaultMappingTemplate.json#L114 

We gotta revisit this to see how better to solve it.


pwnbus:
We decided to remove this field from the default mapping instead (https://github.com/mozilla/MozDef/pull/894), so this PR is no longer needed.

PR #902, opened 2018-10-30
Create an alert on attempts to use known exfil domains

caggle:
For example, pastebin.com


caggle:
I have basically used "proxy_drop_executable.py" alert (and its corresponding unit test) as a template for this one, they are very similar.
One potential caveat I thought of was, if this alert would catch on CONNECTs, as most of these domains in question are available over HTTPS only.

alerts/proxy_drop_exfil_domains.conf
```diff
@@ -0,0 +1,2 @@
+[options]
+exfil_domains = pastebin.com,pasted.co,hastebin.com,dumpz.org,www.pastefs.com,send.firefox.com,gist.github.com
```
claudijd: We should provide less examples here.  I think two is enough.

pwnbus:
Very nice!

PR #905, opened 2018-10-30
Fix geolite db location

pwnbus:


pwnbus:
@zsck How do you build and push a new update to pypi?

PR #906, opened 2018-10-30
Watchlist feature - Ready for Review

Phrozyn:
This is a work in progress to add a watchlist feature that will integrate with our generic alerts.
It will present an interface which you can add specific terms to watch for. Generic alerts will pull those terms and search for them on a schedule until they are removed.

This currently saves the watched term to mongo and to a txt file in the static directory.
There is a specific api endpoint that will be protected by nginx and psk in the form of a jwt that the alert engine will call.
Everything tested to be working.

Just correcting pep8 stuff now


rest/index.py
```diff
@@ -72,6 +72,21 @@ def status():
     return response
 
 
+# placeholder for future watchlist request #
```
pwnbus: Is this comment still relevant?

rest/index.py
```diff
@@ -521,6 +545,44 @@ def kibanaDashboards():
     return json.dumps(resultsList)
 
 
+def getWatchlist():
+    WatchList = []
```
pwnbus: There's some inconsistencies for this variable name. You define it here with a capital 'W', but then later on, you reference it using the lowercase version. I think we should use the lowercase variable name throughout.

rest/index.py
```diff
@@ -72,6 +72,21 @@ def status():
...
+@route('/getwatchlist/')
+def status():
+    '''endpoint for grabbing watchlist contents'''
+    if request.body:
+        request.body.read()
+        request.body.close()
+    response.status = 200
+    response.content_type = "application/json"
+    response.body = getWatchlist()
+    sendMessgeToPlugins(request, response, 'getwatchlist')
```
pwnbus: This 'sendMessageToPlugins' line isn't needed, as we don't actually have a plugin that registers for "getwatchlist". I believe you can remove the line entirely.

alerts/get_watchlist.py
```diff
@@ -0,0 +1,93 @@
...
+from requests_jwt import JWTAuth
+import sys
+
+logger = logging.getLogger()
+formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+
+
+class AlertWatchList(AlertTask):
+    def main(self):
+        global watchterm
```
pwnbus: Instead of using global variables here, you can use an instance variable to keep track of the watchterm. For example `self.watchterm = 'someusername'`

rest/index.py
jeffbryner: It would fit the pattern better if the thing that makes up the watchlist is actually a plugin that could be custom to one's environment, but gets called via the rest interface. i.e. like the blocklist, or logincounts, etc. The endpoint is generic but the plugin has the business logic/tech integration/etc.

jeffbryner: oh, this is the 'get' endpoint.. /me reads more

rest/plugins/watchlist.py
```diff
@@ -0,0 +1,159 @@
...
+                if 'userid' in i.keys():
+                    userid = i.values()[0]
+
+            if watchitem and watchcontent is not None:
+                if len(watchcontent) < 2:
+                    sys.stderr.write('{0} does not meet requirements. Not added. \n'.format(watchcontent))
+
+                else:
+                    watchlisted = False
+                    if watchlisted is False:
```
jeffbryner: Is watchlisted ever not False?

requirements.txt
```diff
@@ -48,6 +48,7 @@ pynsive==0.2.6
 pyparsing==2.1.10
 python-dateutil==2.6.1
 pytz==2017.3
+requests-jwt==0.5.3
```
jeffbryner: This is a bit concerning: 
https://github.com/tgs/requests-jwt/blob/master/README.rst#this-package-is-not-actively-maintained-and-has-never-had-a-security-audit

alerts/get_watchlist.py
```diff
@@ -0,0 +1,93 @@
...
+                    user = ev['details']['username']
+                    user_data = 'by {}'.format(user)
+                elif 'originaluser' in ev['details']:
+                    user = ev['details']['originaluser']
+                    user_data = 'by {}'.format(user)
+                elif 'user' in ev['details']:
+                    user = ev['details']['user']
+                    user_data = 'by {}'.format(user)
+                else:
+                    user_data = 'by an unidentified user'
```
jeffbryner: rather than set these fallbacks at the end of a series of if/elses, I find it better to set defaults up front so you never run into a scenario where you have an undefined variable. I'd advocate for continuing the series of setting defaults just after line 63 to include user_data, source_data and anything else that should have a default, then just override the default in the if/then series based on what you find.

meteor/client/addwatchItem.js
```diff
@@ -0,0 +1,63 @@
...
+                        if ( result == true){
+                            Session.set('displayMessage','listed watch item successfully ');
+                        }else{
+                            Session.set('errorMessage','watch item creation failed, returned & ' + JSON.stringify(result) );
+                        }
+                    }else{
+                        Session.set('errorMessage','watch item failure & ' + JSON.stringify(err));
+                    }
+                });
+            Router.go('/watchitem');
```
jeffbryner: I think you want watchlist instead of watchitem unless you supply a term watchitem/term.

meteor/client/watchlistTable.js
```diff
@@ -0,0 +1,40 @@
...
+            return watchlist.find({},{
+                                   sort: {dateExpiring: -1}
+                                });
+        }
+    });
+
+    //select an incident for editing
+    Template.watchlist.events({
+        "click .watchedadd": function(e,t){
+            //clear any leftover ip session val
```
jeffbryner: s/ip//

rest/index.py
jeffbryner: Still probably a good idea to pull the getWatchlist() function out into a plugin. That would give folks who don't want the UI a way to still use the watchlist feature fed by something else.

rest/index.py
Phrozyn: ha, no, will revise

rest/index.py
Phrozyn: these are two different variables. index.py uses python based WatchList variable for it's database results. The other is node based variables, which do slightly different things.

rest/index.py
Phrozyn: We have a way to use this without the UI, it's called generic alerts. 
The point of this is to have the ability to quickly add a watchlist item through the UI without having to create PR or wait for a code update.

rest/index.py
pwnbus: I'm specifically referring to https://github.com/mozilla/MozDef/pull/906/files#diff-c9dff09105e9694576c5a91b26bcd704R549 and https://github.com/mozilla/MozDef/pull/906/files#diff-c9dff09105e9694576c5a91b26bcd704R554 (both of which are in index.py).

rest/plugins/watchlist.py
Phrozyn: This can definitely be improved, I removed the logic that determined if an item is watchlisted, that's handled earlier in this code. So we could essentially remove the check for is false and simply input the data.

alerts/get_watchlist.py
Phrozyn: Thanks! Good advice!

requirements.txt
Phrozyn: the newest version was released about a month ago, and was updated 4 months ago. I can reach out to the maintainer and ask if that banner is old cruft from when it wasn't maintained.

alerts/get_watchlist.py
Phrozyn: Will try that, thank you!

requirements.txt
Phrozyn: So the maintainer says he only added that to say he's not actively using this project anymore, and isn't working with jwt anymore, but he would provide minor fixes if needed.
This particular package is doing a very small piece of what we need (taking a secret and providing a hash of it in conjunction with "requests" lib.

I'd have to see what else is out there before ripping it out  to replace this functionality.

rest/index.py
Phrozyn: I've modified this in a way that these two vars no longer imitate one another

requirements.txt
Phrozyn: Per our conversation in the meeting today, we will move forward with this, and I'll work with Andrew to implement an even better solution utilizing auth0 authorization mechanism for machine to machine communication.

PR #913, opened 2018-10-31
Add templater for alert file creation

claudijd:


claudijd:
Going to pair with pwnbus on this and make sure it works with make strategy

PR #916, opened 2018-10-31
Add ElasticSearch service linked IAM role

### LINKED ISSUE #787: Cloudy MozDef Cloudformation Does not Create Managed ES Service Linked Role
Cloudformation can't form the es service linked role.  
```
aws iam create-service-linked-role --region {{ aws_region }} --aws-service-name es.amazonaws.com --description ansibled-es-role

```

gene1wood:
Also pull secrets out of a JSON file and into an environment variable. Fixes #787


cloudy_mozdef/Makefile
```diff
@@ -8,7 +8,8 @@ STACK_PARAMS	:= file://aws_parameters.json
 S3_BUCKET_NAME  := mozdef.infosec.allizom.org
 S3_BUCKET_PATH	:= cf
 S3_BUCKET_URI	:= s3://$(S3_BUCKET_NAME)/$(S3_BUCKET_PATH)
-S3_STACK_URI	:= https://s3-$(AWS_REGION).amazonaws.com/$(S3_BUCKET_NAME)/$(S3_BUCKET_PATH)/mozdef-parent.yml
+S3_STACK_URI	:= https://s3-$(AWS_REGION).amazonaws.com/$(S3_BUCKET_NAME)/$(S3_BUCKET_PATH)/
+# OIDC_CLIENT_SECRET is set in an environment variable by running ". aws_parameters.sh"
```
gdestuynder:
```suggestion
# OIDC_CLIENT_SECRET is set in an environment variable by running "source aws_parameters.sh"
```

gdestuynder: clarity

gdestuynder:
@gene1wood  is the env variable from a file better than just a file? in both case there's a file, so i'm wondering. Is this so that you can import it from another directory than the example?

If so i think i wouldn't include and example because it makes it likely to be wrongly committed with a secret

cloudy_mozdef/Makefile
gene1wood: Agreed!

gene1wood:
@gdestuynder The driver for the env variable instead of json file is so we can [include the template location on the command line](https://github.com/mozilla/MozDef/pull/916/files#diff-a3981a89b8eb06bdd63d0d2f600573edR31). Without this (or some other solution), when you use the makefile to upload to a development area in S3, that location isn't then consumed when you create or update the stack because the parameter isn't passed in.

> is the env variable from a file better than just a file? in both case there's a file,

This isn't intended to address the problem we'd talked about the other day of the secrets being in the CWD of the git repo. So no, an env file isn't better than a json file in regards to the downsides of having the secrets in the git repo.

> If so i think i wouldn't include and example because it makes it likely to be wrongly committed with a secret

I included an `aws_parameters.example.sh` file and updated `.gitignore` to ensure `aws_parameters.sh` doesn't get committed. Does that work or should we do something more or different?

gdestuynder:
@gene1wood I'm thinking we could just document the fact that you should have it in the env and not provide the file - i think a lot of people just copy paste in their env from their secret manager, while if its in a .sh file they'll leave it in the repo anyway and forget (just like the json file)

gene1wood:
Agreed, I'll open an issue so we can move away from a file model and to something safer/better.

cloudy_mozdef/dmake
```diff
@@ -37,4 +37,5 @@ exec docker run --rm --name ${CONTAINER_NAME} \
   -e "AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}" \
   -e "AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}" \
   -e "AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}" \
+  -e "OIDC_CLIENT_SECRET=${OIDC_CLIENT_SECRET}"
```
gene1wood: Whoops, missing a trailing `\`

PR #919, opened 2018-11-01
Enable dmake to use all AWS environment variables

gene1wood:
In #916 there was a missing `\` character [here](https://github.com/mozilla/MozDef/pull/916#pullrequestreview-170814992).

This PR fixes that and tries to accommodate all AWS environment variables.

I've tried testing this and it looks correct. When I bash into the container
* the environment variables are present
* the mounted files are mounted in the locations that the environment variables point to

### DIFF
--- a/cloudy_mozdef/dmake
+++ b/cloudy_mozdef/dmake
@@ -39,9 +39,7 @@ fi
 
 dmake_env_file="`mktemp`"
 trap "{ rm -f \"$dmake_env_file\"; }" EXIT
-env | egrep "^AWS" > "$dmake_env_file"
-
-cat $dmake_env_file
+env | egrep "^AWS|OIDC_CLIENT_SECRET" > "$dmake_env_file"
 
 exec docker run --rm --name ${CONTAINER_NAME} \
   -u $(id -u) \
@@ -49,5 +47,4 @@ exec docker run --rm --name ${CONTAINER_NAME} \
   ${config_file_mount} \
   -v $(pwd):${DOCKER_PROJECT_DIR} \
   --env-file "$dmake_env_file" \
-  -e "OIDC_CLIENT_SECRET=${OIDC_CLIENT_SECRET}" \
   ${HUB}/${IMG_NAME}:latest make $@
--- a/cloudy_mozdef/dmake
+++ b/cloudy_mozdef/dmake
@@ -29,13 +29,25 @@ function check_img() {
 
 check_img || exit 127
 
+if [ -n "$AWS_CONFIG_FILE" ]; then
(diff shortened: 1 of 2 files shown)


cloudy_mozdef/dmake
```diff
@@ -29,13 +29,25 @@ function check_img() {
...
+
+if [ -n "$AWS_SHARED_CREDENTIALS_FILE" ]; then
+  config_file_mount="${config_file_mount} -v ${AWS_SHARED_CREDENTIALS_FILE}:${AWS_SHARED_CREDENTIALS_FILE}"
+fi
+
+dmake_env_file="`mktemp`"
+trap "{ rm -f \"$dmake_env_file\"; }" EXIT
+env | egrep "^AWS" > "$dmake_env_file"
+
+cat $dmake_env_file
```
gdestuynder:
```suggestion
echo "Will use the following environment variables:"
cat $dmake_env_file
```

cloudy_mozdef/dmake
```diff
@@ -29,13 +29,25 @@ function check_img() {
...
   -v ${AWS_CREDS_DIR}:/root/.aws \
+  ${config_file_mount} \
   -v $(pwd):${DOCKER_PROJECT_DIR} \
-  -e "AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}" \
-  -e "AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}" \
-  -e "AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}" \
-  -e "AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}" \
-  -e "OIDC_CLIENT_SECRET=${OIDC_CLIENT_SECRET}"
+  --env-file "$dmake_env_file" \
+  -e "OIDC_CLIENT_SECRET=${OIDC_CLIENT_SECRET}" \
```
gdestuynder: im just assuming this doesnt work in the env file, or that we fear people would hard code it?

cloudy_mozdef/dmake
gene1wood: Doh! Sorry, that was what I was using to debug and I left it in. I'll just remove it. Thanks for catching that.

cloudy_mozdef/dmake
gene1wood: We could either pass OIDC_CLIENT_SECRET to docker by appending it to the temporary file containing the environment variables or via this command line argument.

I guess it comes down to whether we prefer the secret to be on disk or on the command line. On disk, if something goes wrong and the temp file doesn't get deleted it's left here. On the command line and it's visible to other processes.

I would say it's probably better to put it in the environment temp file. What do you think @gdestuynder ?

PR #920, opened 2018-11-01
Update docs to style commands

pwnbus:


pwnbus:
screen shot 2018-11-01 at 5 30 50 pm

This styles the commands to look like the following

PR #934, opened 2018-11-05
A MozDef plugin that parses GitHub's Webhook events to create meaning…

mpurzynski:
…ful IR data


mpurzynski:
Oops, forgot to push the file with unit tests. You're welcome. 60 LoC, 3000 lines of tests.

mpurzynski:
I refuse to spend several days "fixing" unit test so intentions are what pep wants them to be. Checks like this on 3000 lines of a unit test full of JSON serve no purpose.

pwnbus:
@mpurzynski Adherence to coding standards promotes openness, readability, and simplicity. We've been lax on these before but it's worth the time investment to make this consistent across the project, no matter if it's tests or the implementation.

Despite the length of the code, it only took 5 minutes to update this code to conform to PEP8 standards, which I've done for you this one time at https://github.com/mpurzynski/MozDef/pull/1.

Let me know if you want and we can figure out a way to make your dev environment as easy to manage for PEP8 as possible (e.g. auto correction, linting). For example, sublime text can be configured to report on pep8 failures when you're typing, as well as autofix certain issues when you save the file.

mpurzynski:
Per request I moved the mapping file to the plugin sub-directory.

mq/esworker_github_webhooks.conf
```diff
@@ -0,0 +1,10 @@
+[options]
```
pwnbus: You don't need to add this file.

mq/plugins/github_webhooks.py
```diff
@@ -0,0 +1,64 @@
...
+
+class message(object):
+    def __init__(self):
+        '''
+            Plugin used to fix object type discretions with cloudtrail messages
+        '''
+        self.registration = ['githubeventsqs']
+        self.priority = 10
+
+        with open('mq/plugins/github_mapping.yml', 'r') as f:
```
pwnbus: We want to get away from using relative paths here, so we should instead do something like:
```
os.path.join(os.path.dirname(__file__), 'github_mapping.yml')
```

mpurzynski: Done

mpurzynski:
Ansible changes merged

PR #945, opened 2018-11-06
Feature flagging

### LINKED ISSUE #936: Feature flagging
#### ISSUE TYPE
 - Feature Idea

#### COMPONENT NAME
Meteor/Mozdef

#### CONFIGURATION
Will add capability to do feature flagging to do things like: 
- turn off block IP (already exists)
- turn off the ip blocklist
- turn off the fqdn blocklist
- turn off individual vizualizations

#### DESCRIPTION
Currently it's an all or nothing experience, but maybe folks just want to use Mozdef to track investigations/incidents, or maybe they just want alerts, or just visualizations. This will allow folks to turn on and off the bits they want. 
This will start with the UI first, then expand to other components if possible.

jeffbryner:
Adds feature flagging, mostly to the UI but in a way that can be extended to other components in future work.

closes #936


docker/compose/mozdef_meteor/files/settings.js
```diff
@@ -12,7 +12,6 @@ mozdef = {
   port: "80",
```
pwnbus: It's worth mentioning here that this file doesn't actually get used anymore in docker meteor environment (we write to the wrong location). This never resulted in errors because we have docker settings in the real settings.js file (for example, we point to the rest api container).

pwnbus: I think we have a couple options:
1. Convert that file over to an environment variable file (similar to how it's done in cloudy mozdef) and then include that .env file in docker-compose.
2) Sync entire contents of https://github.com/mozilla/MozDef/blob/master/meteor/imports/settings.js with the file in the meteor container, however, that settings file now contains meteor specific logic such as https://github.com/mozilla/MozDef/blob/master/meteor/imports/settings.js#L31-L33, so it would be preferable to decouple https://github.com/mozilla/MozDef/blob/master/meteor/imports/settings.js#L10-L18 from the settings file in meteor, so we only overwrite the settings dict and nothing else. (We have a full version of that file with all of the meteor stuff in production, so we could use this functionality regardless)

pwnbus: I think if we remove the file, then we're now saying "in order for docker to run, the meteor/imports/settings.js file must be using docker environment specific values" (which is what it currently is doing unless you use environment vars like cloudy mozdef) but I think it's preferable to keep the two configurations not dependent on each other.

jeffbryner: Probably rm the file and move to env variables is the right choice rather than having a one off file/source that we copy around.

pwnbus: Agreed!

jeffbryner: OK, added some sanity checking for bool environment variables and cleaned up the way this all interfaces with the accounts system.

docs/source/advanced_settings.rst
```diff
@@ -16,6 +16,23 @@ These files all follow the same format:
...
+
+::
+
+    OPTIONS_METEOR_ROOTURL is "http://localhost" by default and should be set to the dns name of the UI where you will run MozDef
+    OPTIONS_METEOR_PORT is 80 by default and is the port on which the UI will run
+    OPTIONS_METEOR_ROOTAPI is http://rest:8081 by default and should resolve to the location of the rest api
+    OPTIONS_METEOR_KIBANAURL is http://localhost:9090/app/kibana# by default and should resolve to your kibana installation
+    OPTIONS_METEOR_ENABLECLIENTACCOUNTCREATION is true by default and governs whether accounts can be created
+    OPTIONS_METEOR_AUTHENTICATIONTYPE is meteor-password by default and can be set to oidc to allow for oidc authentication
+    OPTIONS_REMOVE_FEATURES is empty by default, but if you pass a comma separated list of features you'd like to remove they will no longer be available.
```
pwnbus: To align with the other environment variables above, should we make this `OPTIONS_METEOR_REMOVE_FEATURES`?

meteor/client/menu.html
```diff
@@ -12,40 +12,64 @@
...
-                </li>
-                <li class="first"><a href="#">Visualizations</a>
-                    <ul>
-                        <li class="first"><a href="/logincounts/">logincounts</a></li>
-                        <li><a href="/attackers">attackers</a></li>
-                        <li class="last"><a href="/globe">globe</a></li>
-                    </ul>
-                </li>
-                <li><a href="/about">About</a></li>
+                {{#if isFeature "kibana"}}
```
pwnbus: One weird thing that I've noticed, is if you turn off a few features via the environment variable, and then rebuild meteor, everytime you refresh the page (or originally load it), the disabled features links originally appear, and then disappear. 

Is there a way to fix this?

docs/source/advanced_settings.rst
jeffbryner: My intention is to have this be the way to disable features system wide (eventually). i..e either passing in a common env variable or having all components connect to the resulting mongo features table to see if they should be enabled. So for now I think it makes sense to leave it generic.

meteor/client/menu.html
jeffbryner: I fought that for a while. The issue is the subscription to the features table is ready before records show up (this is universal to all collections and is just the async way meteor works). Since the 'menu' template is reactive it removes the features once the data shows up; but initially it's empty so everything shows. 
I'll keep looking for more elegant ways to avoid showing the menu until we have records; but I don't know of one currently.

jeffbryner: Just had a brainstorm and added a utility function the template can use to know when records are ready. It now delays painting until it knows the features.

PR #947, opened 2018-11-08
The standard alert version of the Address_Scan

mpurzynski:


alerts/nsm_scan_address.json
```diff
@@ -0,0 +1,4 @@
+{
+    "sourcemustmatch":"[10.0.0.0 TO 10.255.255.255]",
+    "sourcemustnotmatch":"10.33.44.54 OR 10.88.77.54 OR 10.76.54.54 OR 10.251.30.138 OR 10.54.65.234"
```
pwnbus: Since we're doing it in python now, would it make sense to break this out so that sourcemustnotmatch is an array of ips to exclude? or do we need the ability to do "complex" 'and' with 'or' statements?

pwnbus: Works for me

PR #954, opened 2018-11-09
Change Meteor Kibana URL to relative

gene1wood:


gene1wood:
Travis failed because of a network timeout to centos yum
```
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
12: Timeout on http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container: (28, 'Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds')
```

PR #959, opened 2018-11-12
Logincounts fix

jeffbryner:
Update the logincounts function to recent versions of events/endpoints, etc. 

This targets the current auth0 events, instead of the historical ldap events.

Ideally 'authentication' events are generic with category: authentication, details.success as true/false, details.username as the userid in question, etc. 

This PR is a first step in making this count/visualization generic.

### DIFF
--- a/rest/index.conf
+++ b/rest/index.conf
@@ -1,5 +1,5 @@
 [options]
 kibanaurl=http://localhost:9090/app/kibana
-esservers=http://localhost:9200
+esservers=http://elasticsearch:9200
 mongoport=3002
 listen_host=0.0.0.0
--- a/rest/index.conf
+++ b/rest/index.conf
@@ -1,5 +1,5 @@
 [options]
 kibanaurl=http://localhost:9090/app/kibana
-esservers=http://elasticsearch:9200
+esservers=http://localhost:9200
 mongoport=3002
 listen_host=0.0.0.0
--- a/rest/plugins/auth0_logincounts.py
+++ b/rest/plugins/auth0_logincounts.py
@@ -14,6 +14,7 @@ from mozdef_util.query_models import SearchQuery, TermMatch, TermsMatch, QuerySt
 from mozdef_util.utilities.toUTC import toUTC
 from mozdef_util.utilities.logger import logger, initLogger
 
+
 class message(object):
     def __init__(self):
         '''register our criteria for being passed a message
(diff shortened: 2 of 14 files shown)


rest/index.conf
```diff
@@ -1,5 +1,5 @@
 [options]
 kibanaurl=http://localhost:9090/app/kibana
-esservers=http://localhost:9200
+esservers=http://elasticsearch:9200
```
pwnbus: I don't know how I feel about us modifying main config files to work in docker (there were a couple other ones recently as well).

I think we should keep these files as they would be for a non docker environment, and then make local modifications if we need it to work in docker, or have docker overwrite these files accordingly.

jeffbryner: Yeah, good point. I didn't realize the rest docker container had it's own .conf. For whatever reason it wasn't working for me without this change, but I'd guess it's something in my local docker setup. I'll revert this.

rest/plugins/auth0_logincounts.py
```diff
@@ -0,0 +1,134 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import requests
```
pwnbus: There are unused imports here, for example:
```
requests
ElasticsearchInvalidIndex
TermsMatch
logger
initLogger
```

rest/plugins/auth0_logincounts.py
```diff
@@ -0,0 +1,134 @@
...
+        '''
+
+        self.registration = ['logincounts']
+        self.priority = 5
+        self.name = "auth0LoginCounts"
+        self.description = "count failures/success logins"
+
+        # set my own conf file
+        # relative path to the rest index.py file
+        self.configfile = './plugins/auth0_logincounts.conf'
```
pwnbus: You reference a config file here, but there isn't one included as part of the PR?

pwnbus: If we're not using the config file anywhere, nor are we using the 'ignoredusers' key from a config file, can we just remove the logic to load a config file entirely?

jeffbryner: we can, but I'll bet we add it right back in the next revision depending on what sort of data shows up. In some cases it seems like auth0 is sending " " as a username for example.

pwnbus: I think that's fine, we can cross that bridge when we get there.

PR #964, opened 2018-11-14
Lower keys

Phrozyn:
This effectively lowers all keys in any event sent through the MQ service.


mq/plugins/lower_keys.py
```diff
@@ -0,0 +1,31 @@
...
+        takes an incoming message
+        and sets the keys to lowercase
+        '''
+
+        self.registration = ['summary']
+        self.priority = 5
+
+    def onMessage(self, message, metadata):
+        def renameKeysToLower(message):
+            if isinstance(message, dict):
```
jeffbryner: I think this isn't going to traverse into really complex sub structures like are present in cloudtrail? (lists of dicts of lists, etc)
https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/dict2List.py is an example of some of the complexity involved in this.

Phrozyn: It does, I have tested it on actual cloudtrail events.  Thanks for the inquiry!

PR #967, opened 2018-11-14
fix up field names, missing field logic closes #958

### LINKED ISSUE #958: Error is always true
https://github.com/mozilla/MozDef/blob/95be0e40eb2f45a43ab4ac77a9fa13117fb4824e/cron/auth02mozdef.py#L320

This line sets error=True in the auth02mozdef.py cron. Looking at events sent via this, error is always true which makes me think the try/except blocks aren't working as planned.

jeffbryner:
Changes to auth02mozdef to fix up field logic to match current logs and better conform to field naming. Most importantly, fixes the details.error always being true, sets specific categories and details.success based on authentication results.

### DIFF
--- a/cron/auth02mozdef.py
+++ b/cron/auth02mozdef.py
@@ -12,6 +12,8 @@ import sys
 import os
 import requests
 import mozdef_client as mozdef
+from mozdef_util.utilities.dot_dict import DotDict
+
 try:
     import urllib.parse
     quote_url = urllib.parse.quote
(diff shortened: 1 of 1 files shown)


cron/auth02mozdef.py
```diff
@@ -271,34 +264,91 @@ def process_msg(mozmsg, msg):
...
     details = DotDict({})
 
+    # key words used to set category and success/failure markers
+    authentication_words=['Login','Logout','Auth']
+    authorization_words=['Authorization','Access','Delegation']
+    success_words=['Success']
+    failed_words=['Failed']
+
+    # default category (might be modified below to be more specific)
+    mozmsg.category='iam'
```
Phrozyn: Just a nitpick, before

cron/auth02mozdef.py
```diff
@@ -308,45 +358,24 @@ def process_msg(mozmsg, msg):
...
-        except KeyError:
-            pass
-
     # Differenciate auto login (session cookie check validated) from logged in and had password verified
 
     try:
         for i in msg.details.prompt:
             # Session cookie check
             if i.get('name') == 'authenticate':
-                details['auth_type'] = 'Login succeeded due to a valid session cookie being supplied'
```
pwnbus: I could be the minority here, but I actually prefer to have the key names contain the '_' to separate words like this, it seems more legible for me.

jeffbryner: Not impossible, but if we wanted to be consistent we would have to change the standard so things like sourceipv4address become source_ip_v4_address, destination_port, source_port, etc. Probably a discussion for another issue/pr.

Issue #871, opened 2018-10-24
Enable pointing Route53 domain name alias at AWS ALB


gene1wood:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html

gene1wood:
We should do this through a feature flag in the parent template which spins a child template that creates the route53 record

PR #975, opened 2018-11-24
Add GuardDuty findings collection to AWS MozDef

### LINKED ISSUE #925: Pass good ROOTURL to meteor
The ROOTURL variable in meteor isn't being set correctly. It should be set through env variables to something like this.

` OPTIONS_METEOR_ROOTURL = https://cloudymozdef.security.allizom.org/`

We'll need to somehow know what domain name is being used to access the site which means this may depend on #871

andrewkrug:
* Make threshold for attackers and cidr blocks configurable.
* Update CloudFormation to configure alert workers
* Add @jeffbryner changes to add guarduty basic alert for port probes
* Set Meteor ROOTURL (Closes #925)
* Add GuardDuty alerts

### DIFF
--- a/cloudy_mozdef/cloudformation/mozdef-instance.yml
+++ b/cloudy_mozdef/cloudformation/mozdef-instance.yml
@@ -71,6 +71,9 @@ Parameters:
   MozDefSQSQueueName:
     Type: String
     Description: The name of the generic SQS queue used to pickup events.
+  DomainName:
+    Type: String
+    Description: The fully qualified domain you'll be hosting MozDef at.
 Resources:
   MozDefElasticLoadBalancingV2TargetGroup:
     Type: AWS::ElasticLoadBalancingV2::TargetGroup
@@ -119,7 +122,8 @@ Resources:
             - content: |
                 OPTIONS_ESSERVERS=${ESURL}
                 OPTIONS_KIBANAURL=${KibanaURL}
-                OPTIONS_METEOR_KIBANAURL=https://relative:9090/_plugin/kibana/
+                OPTIONS_METEOR_KIBANAURL=https://relative:9090/_plugin/kibana/ # not a typo needed for meteor
+                OPTIONS_METEOR_ROOTURL=https://${DomainName}
                 # See https://github.com/mozilla-iam/mozilla.oidc.accessproxy/blob/master/README.md#setup
                 client_id=${OIDCClientId}
                 client_secret=${OIDCClientSecret}
@@ -136,8 +140,10 @@ Resources:
                 cookiename=sesmeteor
                 # Increase the AWS ES total fields limit from 1000 to 4000
                 OPTIONS_MAPPING_TOTAL_FIELDS_LIMIT=4000
+                # Set thresholds for attack dataviz lower means more ogres
+                OPTIONS_IPV4ATTACKERHITCOUNT=5
+                OPTIONS_IPV4ATTACKERPREFIXLENGTH=24
               path: /opt/mozdef/docker/compose/cloudy_mozdef.env
-          write_files:
             - content: |
                 #!/usr/bin/env python
                 # This Source Code Form is subject to the terms of the Mozilla Public
@@ -152,7 +158,7 @@ Resources:
                 ALERTS = {
                     'bruteforce_ssh.AlertBruteforceSsh': {'schedule': crontab(minute='*/1')},
                     'unauth_ssh.AlertUnauthSSH': {'schedule': crontab(minute='*/1')},
-                    'guard_duty_probe.AlertGuardDutyProbe': {'schedule': crontab(minute='*/1')}',
+                    'guard_duty_probe.AlertGuardDutyProbe': {'schedule': crontab(minute='*/1')},
                     'cloudtrail_logging_disabled.AlertCloudtrailLoggingDisabled': {'schedule': timedelta(minutes=1)},
                     'cloudtrail_deadman.AlertCloudtrailDeadman': {'schedule': timedelta(hours=1)}
                 }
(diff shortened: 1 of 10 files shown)


andrewkrug:
This duplicates #974 and #975 by @jeffbryner .  If there's no complaint I'd just merge this all at once since it supports each other.

cloudy_mozdef/cloudformation/mozdef-instance.yml
```diff
@@ -71,6 +71,9 @@ Parameters:
   MozDefSQSQueueName:
     Type: String
     Description: The name of the generic SQS queue used to pickup events.
+  DomainName:
```
gene1wood: A note : I was hoping to do this (in #871) by passing in a zone and having CloudFormation handle the route53 record. I may switch to that model down the road.

cloudy_mozdef/cloudformation/mozdef-instance.yml
```diff
@@ -119,7 +122,8 @@ Resources:
             - content: |
                 OPTIONS_ESSERVERS=${ESURL}
                 OPTIONS_KIBANAURL=${KibanaURL}
-                OPTIONS_METEOR_KIBANAURL=https://relative:9090/_plugin/kibana/
+                OPTIONS_METEOR_KIBANAURL=https://relative:9090/_plugin/kibana/ # not a typo needed for meteor
```
gene1wood: I've looked and compared this line that you've added the comment to but can't tell what changed. Was there a missing typo that was re-added? What is this line change about?

cloudy_mozdef/cloudformation/mozdef-instance.yml
```diff
@@ -136,7 +140,86 @@ Resources:
...
+                    'mqserver': 'rabbitmq',
+                    'mquser': 'guest',
+                    'mqpassword': 'guest',
+                    'mqport': 5672,
+                    'alertexchange': 'alerts',
+                    'alertqueue': 'mozdef.alert'
+                }
+
+                ES = {
+                    'servers': ["${ESURL}"]
```
gene1wood: Is it not possible to use the existing `/opt/mozdef/docker/compose/cloudy_mozdef.env` pathway which makes the `ESURL` value available instead of creating config.py?

For example instead of creating `config.py` here in the user-data of the instance, change the `config.py` on disk to something like

```
ES = {
    'servers': [os.environ.get('OPTIONS_ESURL')]
```

cloudy_mozdef/packer/packer.json
```diff
@@ -30,7 +33,7 @@
         "sudo systemctl enable docker",
         "sudo mkdir -p /opt/mozdef/",
         "sudo git clone https://github.com/mozilla/MozDef /opt/mozdef",
-        "cd /opt/mozdef && sudo git checkout guardduty"
+        "cd /opt/mozdef && sudo git checkout reinvent"
```
gene1wood: Do we want this PR in master or should this go in a `reinvent` branch?

tests/alerts/test_guard_duty_probe.py
```diff
@@ -0,0 +1,78 @@
...
+    # alert to trigger
+    default_event = {
+        "_type": "event",
+        "_source": {
+            "source": "guardduty",
+            "details": {
+                "sourceipaddress": "1.2.3.4",
+                "finding": {"action":
+                            {"actionType": "PORT_PROBE"}
+                            }
```
gene1wood: Funny indent

cloudy_mozdef/cloudformation/mozdef-instance.yml
andrewkrug: We should remove my comment line here.  It actually is treated as a string in CF when you do this.  Comments need to be separate lines.

cloudy_mozdef/packer/packer.json
andrewkrug: We should switch this back to master.  I missed this.  Need to come up with a better build / deploy for launching a stack off a branch based build.

tests/alerts/test_guard_duty_probe.py
andrewkrug: Can you just fix this in the github editor?

cloudy_mozdef/cloudformation/mozdef-instance.yml
andrewkrug: We also need to template out that file to activate different alerts.  I'd suggest merging this as is and filing an issue to change the way this works entirely to use a different config system.

cloudy_mozdef/cloudformation/mozdef-instance.yml
andrewkrug: I think assuming that roue53 exists in the customer acct is a big leap.

gene1wood: I was thinking of making it conditional (if you provide a zone as a parameter, use it, otherwise expect the user to deal with names on their own)

cloudy_mozdef/cloudformation/mozdef-instance.yml
gene1wood: @andrewkrug What was the typo you were referring to in the attempted comment?

cloudy_mozdef/packer/packer.json
gene1wood: @andrewkrug Should I change this from `reinvent` to `guardduty` or from `reinvent` to `master`?

tests/alerts/test_guard_duty_probe.py
gene1wood: Done

PR #986, opened 2018-11-28
Add initial alert guide

darakian:
I wrote this guide after working through my first alert and thought it might be helpful to others.

### DIFF
--- a/docs/source/alert_development_guide.rst
+++ b/docs/source/alert_development_guide.rst
@@ -127,7 +127,7 @@ Open the file
 or simply
 ::
 
-  mozdef_alerts/files/config.py
+  alerts/files/config.py
 
 if you are not working from the docker images
 and add your new foo alert to the others with a crontab style schedule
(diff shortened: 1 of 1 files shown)


docs/source/alert_development_guide.rst
```diff
@@ -73,6 +73,74 @@ How to get the alert in MozDef?
...
+Next we will need to enable the log and to schedule it. At time of writing this is a bit annoying.
+Open the file
+::
+
+  docker/compose/mozdef_alerts/files/config.py
+
+or simply
+::
+
+  mozdef_alerts/files/config.py
```
pwnbus: This should be `alerts/files/config.py`

darakian: Gotcha, updated.

darakian:
@pwnbus Where would I find the docs on the differences between PhraseMatch/Term[s]Match/etc...?

pwnbus:
@darakian I've got some docs internal that describe what each one is used for, let me try and add them to the repo

pwnbus:
@darakian I created https://github.com/mozilla/MozDef/pull/988 which I think adds what I originally had on hand, with the expectation that these will get improved in the future.

darakian:
@pwnbus thanks!

PR #993, opened 2018-12-05
The base layer of the read-write alert plugins functionality

mpurzynski:


Phrozyn:
What does this do?

Also you have some conflicts to work out

PR #994, opened 2018-12-05
Proposing fix and a change for proxy alert on exfil domains

caggle:
Earlier I realised that this alert did not fire in production. Thanks to @claudijd I think I identified the root cause (i.e. it was a ES filter regex issue). 

While working on it, I also realised this alert would have only fired on proxy DROPs, and would not fire on other proxy events (e.g. CONNECT for https:// links).

So I decided to remove that filter - currently it's meant to scan on all proxy events for these domains. I have changed the name of the alert accordingly, and updated the tests.


pwnbus:
Very nice!

PR #1003, opened 2018-12-14
Generic logincounts

jeffbryner:


jeffbryner:
this syncs up the logincounts data viz with the normalization standard we've applied to auth0/duo to have category:authentication with details.username and details.success as fields signaling auth success/failures.

PR #1008, opened 2018-12-17
removing unused correlateUserMacAddress files.

Phrozyn:


pwnbus:
I think we can also delete cron/oui.txt as well?

jeffbryner:
Why remove?

Phrozyn:
> Why remove?

It doesn't work. Brandon and I discussed and decided to remove for now until https://trello.com/c/zHNNKzhs/496-if-we-want-mac-addresses-of-users-we-need-to-rewrite-the-cron-for-it gets assigned.

PR #1011, opened 2018-12-18
Add hostname to trace audit alert summary

pwnbus:


tristanweir:
Heh, just came to see how to add this.

PR #1013, opened 2018-12-19
Resolving areas where keys are manipulated after lower_keys is run.

Phrozyn:


mq/plugins/fxaFixup.py
```diff
@@ -77,27 +77,35 @@ def onMessage(self, message, metadata):
                 if message['details']['name'] == 'fxa-auth-server':
                     message['category'] = 'fxa-auth-server'
 
+            if 'fileName' in message['details'].keys():
```
pwnbus: I don't think these statements will ever get triggered, because by the time this plugin runs, all the keys should be lowercased.

Phrozyn: We spoke of this on slack, and came to a solution which has been committed to revert this.

PR #1016, opened 2018-12-22
ipsearch addition to ipmenu dropdown

Phrozyn:
This adds a search link to the ipmenu dropdown to allow an end user to search for the specified IP showing up in an alert with Kibana - kibana is opened up in a new tab and auto-populated.


meteor/client/layout.js
```diff
@@ -44,6 +44,15 @@ if (Meteor.isClient) {
             Session.set('ipintelipaddress',($(e.target).attr('data-ipaddress')));
             $('#modalintelwindow').modal()
         },
+        "click .ipmenu-search": function(e,t){
+            var ipText=$(e.target).attr('data-ipaddress');
+            console.log("IP: " + ipText)
+            var searchDomain=getSetting('kibanaURL');
+            var searchPath="#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-1h,mode:quick,to:now))&_a=(columns:!(_source),index:AWdcCjsPo9lh6BmjcvcD,interval:auto,query:(query_string:(analyze_wildcard:!t,query:'details.sourceipaddress:"+ipText+"')),sort:!(utctimestamp,desc))";
```
pwnbus: This does not work as the index hash "AWdcCjsPo9lh6BmjcvcD" is specific to each kibana installation. (meaning this won't always be the hash of the index pattern name)

Phrozyn: good point.

Phrozyn: I've replaced with the plaintext version of "events-weekly"
Also asked you to automagically populate the id of index-patterns so we don't have to deal with hashes, this should resolve the issue :) 

Thanks!

rest/index.py
```diff
@@ -178,6 +178,16 @@ def ipintel():
     return response
 
 
+@route('/ipsearch', methods=['POST'])
+@route('/ipsearch/', methods=['POST'])
+def ipsearch():
```
pwnbus: We don't need to define this route, meteor handles the opening of the url so this is never actually called.

Phrozyn: hmm, I think  you are right, originally I was using this section to open the url, but later moved it.
I'll update :) ty!

PR #1018, opened 2018-12-26
Update slack bot

pwnbus:


bot/slack/commands/ip_info.py
```diff
@@ -0,0 +1,49 @@
...
+
+def is_ip(ip):
+    try:
+        netaddr.IPNetwork(ip)
+        return True
+    except Exception:
+        return False
+
+
+def ipLocation(ip):
```
arcrose: For consistency, let's rename this to `is_location` as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

bot/slack/commands/ip_info.py
```diff
@@ -0,0 +1,49 @@
...
+    location = ""
+    try:
+        geoip_data_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../../data/GeoLite2-City.mmdb")
+        geoip = GeoIP(geoip_data_dir)
+        geoDict = geoip.lookup_ip(ip)
```
arcrose: I think we should really avoid falling into the habit of mixing `snake_case` and `camelCase` as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

bot/slack/commands/ip_info.py
```diff
@@ -0,0 +1,49 @@
...
+            location = geoDict['country_name']
+            if geoDict['country_code'] in ('US'):
+                if geoDict['metro_code']:
+                    location = location + '/{0}'.format(geoDict['metro_code'])
+    except Exception:
+        location = ""
+    return location
+
+
+class command():
```
arcrose: Class names should always be in `PascalCase` as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#class-names)

bot/slack/commands/ip_info.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self):
+        self.command_name = '!ipinfo'
+        self.help_text = 'Perform a geoip lookup on an ip address'
+
+    def handle_command(self, parameters):
+        response = ""
+        for ip_token in parameters:
+            if is_ip(ip_token):
+                ip = netaddr.IPNetwork(ip_token)[0]
+                if (not ip.is_loopback() and not ip.is_private() and not ip.is_reserved()):
```
arcrose: Is it not considered bad style to put conditions in parens when they don't have to be?
This condition _could_ be shortened from

```py
if not A and not B and not C
```

to

```py
if not (A or B or C)
```

as per [DeMorgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

bot/slack/commands/ip_whois.py
```diff
@@ -0,0 +1,31 @@
+import netaddr
+from ipwhois import IPWhois
+
+
+def is_ip(ip):
```
arcrose: We should prefer to move shared code to a module where it can be imported in both cases. Unnecessary duplication will get us back to a point where we have modified duplicates littered everywhere and make it difficult to refactor.

bot/slack/commands/ip_whois.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        netaddr.IPNetwork(ip)
+        return True
+    except Exception:
+        return False
+
+
+class command():
```
arcrose: As before, class names should be in `PascalCase`.

bot/slack/commands/roulette.py
```diff
@@ -0,0 +1,25 @@
+import random
+
+
+class command():
```
arcrose: As before, class names should be in `PascalCase`.

bot/slack/run_async.py
```diff
@@ -0,0 +1,33 @@
...
+    t2 = task2()
+    ...
+    t1.join()
+    t2.join()
+    """
+    from threading import Thread
+    from functools import wraps
+
+    @wraps(func)
+    def async_func(*args, **kwargs):
```
arcrose: I'm not a fan of the idea of using a decorator here.  This function does not modify its function argument at all, but rather has side-effects, which makes for a leaky and obtuse abstraction.

While requiring more code, we should simply write the three lines of this function's body where calls to an async function occur rather than hiding its behavior this way.

bot/slack/slack_bot.py
```diff
@@ -0,0 +1,161 @@
...
+
+greetings = [
+    "mozdef bot in da house",
+    "mozdef here..what's up",
+    "mozdef has joined the room..no one panic",
+    "mozdef bot here..nice to see everyone"
+]
+
+
+def formatAlert(alert_dict):
```
arcrose: Function names should be in `snake_case` as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

bot/slack/slack_bot.py
```diff
@@ -0,0 +1,161 @@
...
+            response = "\nHelp is on it's way...try these:\n"
+            for plugin in self.plugin_set.enabled_plugins:
+                response += "\n{0} -- {1}".format(
+                    plugin['command_name'],
+                    plugin['help_text']
+                )
+        else:
+            # Loop through plugins and send message to it
+            plugin_command_found = False
+            for plugin in self.plugin_set.enabled_plugins:
```
arcrose: Rather than looping through plugins to find the one we're looking for, we should store all 'registered' plugins in a dictionary contained by the `SlackBot` so that lookup can be done in constant time and more explicitly. We're using the [Command Pattern](https://en.wikipedia.org/wiki/Command_pattern) anyway, so we might as well do it properly.

bot/slack/slack_bot.py
```diff
@@ -0,0 +1,161 @@
...
+                if command == plugin['command_name']:
+                    plugin_command_found = True
+                    response += "\n" + plugin['plugin_class'].handle_command(parameters)
+                    continue
+
+            if plugin_command_found is False:
+                response = "Unknown command: " + command + ". Try !help"
+        return response
+
+    def parse_command(self, content):
```
arcrose: I'm not sure that this approach will serve us well in the long term.

It wouldn't take much more effort to build parsing/tokenizing logic into each command. This way, the SlackBot only has to be able to identify commands by the first "word" in a command string, and can delegate the work of making sense of tokens to the command itself.

bot/slack/slack_bot.py
```diff
@@ -0,0 +1,161 @@
...
+        self.slack_client.api_call(
+            "chat.postMessage",
+            as_user="true",
+            channel=channel,
+            text=text,
+            thread_ts=thread_ts
+        )
+
+    def post_attachment(self, message, channel, color):
+        if channel is None:
```
arcrose: Would it make sense to give the `channel` argument to `post_attachment` a default value of `None` so that callers don't have to specify it in what I assume will be the majority of calls?

bot/README.md
```diff
@@ -1,27 +1,8 @@
...
-Other Resources
----------------
-
-Useful reference documents for those working with the IRC protocol as a client:
-
- * [RFC 2812](http://tools.ietf.org/html/rfc2812)
- * [ISUPPORT draft](http://tools.ietf.org/html/draft-brocklesby-irc-isupport-03)
- * [List of numeric replies](https://www.alien.net.au/irc/irc2numerics.html)
+### IRC
+### Slack
```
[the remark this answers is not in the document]
pwnbus: I combined the IRC and slack Readme into this bot/Readme.md in https://github.com/mozilla/MozDef/pull/1018/commits/ce377adb53addc9d902af988a8fe32851e0714c5

bot/slack/commands/ip_info.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/e05243983cb9167303a19e85a3c88f74da8e2612

bot/slack/commands/ip_info.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/ffb0a3ca5e9c0525e89a8745c064f7273a5a4036

bot/slack/commands/ip_info.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/458723466763d1f51f2879147d4c1c255c4ae221

bot/slack/commands/ip_info.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/f0f30519a71750afa8dea1db5de8f67f622be039

bot/slack/commands/ip_whois.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/458723466763d1f51f2879147d4c1c255c4ae221

bot/slack/commands/roulette.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/458723466763d1f51f2879147d4c1c255c4ae221

bot/slack/run_async.py
pwnbus: This was a copy/pasta from the original IRC mozdefbot, so could you explain this a little more of what you're asking please? I'm honestly not entirely sure what `@wraps` is doing here.

bot/slack/slack_bot.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1018/commits/46b9e7e0650f7032997e07369c724098375fa02a

bot/slack/slack_bot.py
pwnbus: This is actually how it's currently setup, we tokenize the entire slack message into a command + parameters (https://github.com/mozilla/MozDef/pull/1018/files#diff-46fed7650379c8a8432b6efca3b11a91R53 and https://github.com/mozilla/MozDef/pull/1018/files#diff-46fed7650379c8a8432b6efca3b11a91R54). We then identify which plugin (if any) is associated with the command, and simply pass the plugin the parameters (https://github.com/mozilla/MozDef/pull/1018/files#diff-46fed7650379c8a8432b6efca3b11a91R69).

Is this the functionality you were requesting, or am I misunderstanding the comment?

bot/slack/slack_bot.py
pwnbus: The post_attachment function is more of an internal function (which I suppose I can rename with an underscore to make that more obvious. The callers of this function (post_info_message, post_welcome_message etc at https://github.com/mozilla/MozDef/pull/1018/files#diff-46fed7650379c8a8432b6efca3b11a91R145) set a default value of None here, so we're doing this "up the chain".

pwnbus: I renamed the post_attachment function to match the private function naming convention at https://github.com/mozilla/MozDef/pull/1018/commits/6d20570292dfebcfbb18f16ae9f98146f36d6be3

bot/slack/slack_bot.py
pwnbus: Implemented in https://github.com/mozilla/MozDef/pull/1018/commits/30346b03e5adc484378fe4953bb59b8ca8bf3cb8

bot/slack/commands/ip_whois.py
pwnbus: I moved this into mozdef_util at https://github.com/mozilla/MozDef/pull/1018/commits/5f67c2d9eb99ab1a74e8420b6eb08334e9cbf241 . The build will fail because we had to bump the version of mozdef-util in requirements.txt, but I haven't built and pushed a new version of mozdef-util yet, so if you could take a look at the previously mentioned changes and give a thumbs up or down, and I'll then push a new version of mozdef-util to pypi, which should cause the build to then start passing.

bot/slack/slack_bot.py
pwnbus: Oh, upon second thought, do you mean to have each plugin be the one to tokenize the parameters? So if the slack message is "!ipinfo 1.2.3.4 5.6.7.8", the parameter we would pass to the plugin would be "1.2.3.4 5.6.7.8" (instead of ['1.2.3.4', '5.6.7.8'] with how it's currently implemented)?

bot/slack/run_async.py
jeffbryner: I think it was trying to add some threading to work around timing issues in the kitnirc client. If we aren't using that anymore you can probably ditch it

arcrose: I'm basically just saying that instead of writing

```py
@async_run
def function(...):
  # ...
```

we should prefer to write

```py
handler = Thread(target=function, ...)
handler.start()
...
```

directly.

bot/slack/slack_bot.py
arcrose: I see what you mean. My suggestion might not be necessary.

bot/slack/run_async.py
pwnbus: I ended up removing the run_async file and implemented it like you recommended.

PR #1026, opened 2018-12-29
Nav update idea

Phrozyn:
This is a reskin of the mozdef interface that moves the navigation to the left side as a collapsible drawer with dropdowns appearing on hover. 

Colors have been changed to a grey/aqua theme


Phrozyn:
Current: (I just noticed the doc links are missing here)
screen shot 2018-12-29 at 1 26 56 pm

After this commit: (we can change the colors to whatever, this is just an idea) The dc chart has a css file we can modify if we want to change the ring and chart colors.
screen shot 2018-12-29 at 1 28 12 pm

gene1wood:
@Phrozyn What are your thoughts on next steps on this PR it looks like it has some conflicts now so if it's still heading towards a merge it'll need conflict resolution. Or if not, can we close it?

jeffbryner:
> @Phrozyn What are your thoughts on next steps on this PR it looks like it has some conflicts now so if it's still heading towards a merge it'll need conflict resolution. Or if not, can we close it?

@gene1wood  The idea was to merge the broader 'themes' PR (which is done), and then morph this branch to be a theme option. So we can probably keep it open and do that?

Phrozyn:
Closing this as we are changing how this will be implemented. It's implementation will be as a theme option.

gene1wood:
@Phrozyn is the code in the `nav_update_idea` branch still needed or do you have what you need to make a theme option and the branch can be deleted?

gene1wood:
Jeff says "I think we need that branch still to mine from to migrate it to a theme"

PR #1037, opened 2019-01-10
Setup default hosted CloudFormation templates

gene1wood:
* Update the AWS launch button link
* Use the new public S3 buckets for hosting the CloudFormation templates (mozilla/security#21)
* Create a new make target to publish templates to this bucket


gene1wood:
I've used the new make target to deploy the new templates and then created a new stack using the new launch button link.

Issue #1041, opened 2019-01-10
Create index mappings for new install of MozDef using AWS ES


gene1wood:
It looks like when you spin up a new MozDef deployment in AWS, either the `.kibana` index already exists (and so isn't a useful indicator that the indexes need to be configured), or there's some timing issue where kibana is made available before the code in #1022 runs and if a user visits kibana before that point, the `.kibana` index is created and #1022 never runs.

The symptoms of this are that a new install of MozDef in AWS shows this page when the user visits kibana

![kibana-configure-index](https://user-images.githubusercontent.com/1134034/50998384-10479c80-14dc-11e9-8a1c-536fa0eb97c6.png)

PR #1042, opened 2019-01-10
Improve logic to determine if index_patterns are created on setup

pwnbus:
Addresses https://github.com/mozilla/MozDef/issues/1041

### DIFF
--- a/docker/compose/mozdef_bootstrap/files/initial_setup.py
+++ b/docker/compose/mozdef_bootstrap/files/initial_setup.py
@@ -13,6 +13,7 @@ from configlib import getConfig
 import json
 import time
 import os
+import sys
 
 from elasticsearch.exceptions import ConnectionError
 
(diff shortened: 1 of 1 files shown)


docker/compose/mozdef_bootstrap/files/initial_setup.py
```diff
@@ -103,8 +104,15 @@
 if kibana_index_name not in all_indices:
     print "Creating " + kibana_index_name
     client.create_index(kibana_index_name)
+
+time.sleep(1)
```
gene1wood: I'd worry about this not being enough to ensure that the index you just created will be available below. You might want to instead do a try except loop over some period of time with your queries

pwnbus: I added some logic to wait 3 times for the index to be ready at  https://github.com/mozilla/MozDef/pull/1042/commits/18f9b7e9e47b95ee8d718ab87f68c40a4257d4bd

gene1wood: Looks good

Issue #1029, opened 2019-01-04
Alerts not firing


darakian:
Hey all,

Maybe it's the holiday gremlins that got to me, but I've updated my docker instance to 1.35 and I can't get alerts working again. I'm down to the following very simple alert which is not working

```
#!/usr/bin/env python

# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.
# Copyright (c) 2014 Mozilla Corporation


from lib.alerttask import AlertTask
from mozdef_util.query_models import SearchQuery, TermMatch, QueryStringMatch, ExistsMatch, PhraseMatch, WildcardMatch


class AlertHelloworld(AlertTask):
    def main(self):
        # Create a query to look back the last 20 minutes
        search_query = SearchQuery(minutes=20)

        # Add search terms to our query
        search_query.add_must([
            TermMatch('category', 'helloworld'),
       #     ExistsMatch('details.sourceipaddress'),
        ])

        self.filtersManual(search_query)
        # Search aggregations on field 'sourceipaddress'
        # keep X samples of events at most
        self.searchEventsAggregated('details.sourceipaddress', samplesLimit=10)
        # alert when >= X matching events in an aggregation
        self.walkAggregations(threshold=1)

    # Set alert properties
    def onAggregation(self, aggreg):
        # aggreg['count']: number of items in the aggregation, ex: number of failed login attempts
        # aggreg['value']: value of the aggregation field, ex: toto@example.com
        # aggreg['events']: list of events in the aggregation
        category = 'hellocategory'
        tags = ['hello', 'world']
        severity = 'WARNING'
        summary = "My first alert!"

        # Create the alert object based on these properties
        return self.createAlertDict(summary, category, tags, aggreg['events'], severity)
```

with 
```
'helloworld.AlertHelloworld': {'schedule': crontab(minute='*/1')},
```
in my `docker/compose/mozdef_alerts/files/config.py`. I'm sending simple messages to my instance with curl
`curl -v --header "Content-Type: application/json"   --request POST   --data '{"tags": ["test"],"category": "helloworld"}'   http://localhost:8080/events` and they are showing up in Kibana.

Is there something obvious that I'm doing wrong? Also is there some way outside of the UI that lets me determine if an alert is working?

jeffbryner:
[...]

PR #1052, opened 2019-01-15
Alert development docker docs

jeffbryner:
As a follow on to #1029, update the docs and offer options for docker development of alerts and other portions of MozDef.


docker/compose/dev-sampledata.yml
```diff
@@ -6,7 +6,7 @@ services:
       context: ../../
       dockerfile: docker/compose/mozdef_sampledata/Dockerfile
     restart: always
-    command: bash -c '/opt/mozdef/envs/mozdef/examples/demo/sampleevents.sh'
+    command: bash -c 'source /opt/mozdef/envs/python/bin/activate && /opt/mozdef/envs/mozdef/examples/demo/sampleevents.sh'
```
pwnbus: The source command is no longer needed as I've added the virtualenvironment path to the base container.

docker/compose/dev-rest.yml
```diff
@@ -6,7 +6,7 @@ services:
       context: ../../
       dockerfile: docker/compose/mozdef_rest/Dockerfile
     restart: always
-    command: bash -c 'python index.py -c index.conf'
+    command: bash -c 'source /opt/mozdef/envs/python/bin/activate && python index.py -c index.conf'
```
pwnbus: Same here, the source command is no longer needed.

docs/source/alert_development_guide.rst
```diff
@@ -141,15 +137,66 @@ and add your new foo alert to the others with a crontab style schedule
...
+
+
+
+To use a live, iterative environment via docker-compose:
+::
+
+  docker-compose -f docker/compose/docker-compose.yml -f docker/compose/dev-alerts.yml -p mozdef up
+
+This will start up all the containers for a mozdef environment and in addition will allow you an adhoc alerts container to work in that loads the /alerts directory as a volume in the container.
+To run the alert you are developing you will need to edit the alerts/files/config.py file as detailed above to schedule your alert. You will also need to edit it to reference the container environment as follows
```
pwnbus: The correct file to modify is actually 'alerts/lib/config.py', not 'alerts/files/config.py'.

docs/source/alert_development_guide.rst
```diff
@@ -141,15 +137,66 @@ and add your new foo alert to the others with a crontab style schedule
...
+    'servers': ['http://elasticsearch:9200']
+  }
+
+Once you've reference the containers, you can shell into the alerts container:
+::
+  docker exec -it mozdef_alerts_1 bash
+
+Next, source the environment and start celery
+::
+  source /opt/mozdef/envs/python/bin/activate && celery -A celeryconfig worker --loglevel=info --beat
```
pwnbus: You shouldn't need to source once you're in the container anymore.

docs/source/alert_development_guide.rst
jeffbryner: good catch! hence the guide ;-]

PR #1053, opened 2019-01-15
Update cloudtrail mapping plugin

pwnbus:


pwnbus:
Closed issue for now since we're no longer receiving these mapping problems.

PR #1057, opened 2019-01-15
Rename user dict into string keys

pwnbus:


cron/duo_logpull.py
```diff
@@ -44,6 +44,12 @@ def normalize(details):
             else:
                 normalized["success"] = False
         normalized[f] = details[f]
+
+    if 'user' in normalized and type(normalized['user']) is dict:
```
andrewkrug: You should use isinstance(normalized['user'], dict) instead of type based comparisons if possible.

mpurzynski: We found out that isinstance() is painfully slow and been trying to migrate away from it, as far as I know.

andrewkrug: Do you know why?

PR #1065, opened 2019-01-23
Squid parsing

mpurzynski:


requirements.txt
```diff
@@ -61,3 +61,4 @@ uritemplate==0.6
 urllib3==1.23
 uwsgi==2.0.17.1
 virtualenv==1.11.4
+tldextract
```
pwnbus: Do you have a specific version that we can pin to?

mq/plugins/squidFixup.py
```diff
@@ -0,0 +1,141 @@
...
+                newmessage[u'details'][u'proto'] = proto
+                newmessage[u'details'][u'mimetype'] = tokens[14]
+                newmessage[u'utctimestamp'] = (toUTC(float(tokens[0])) - timedelta(milliseconds=float(tokens[1]))).isoformat()
+                newmessage[u'timestamp'] = (toUTC(float(tokens[0])) - timedelta(milliseconds=float(tokens[1]))).isoformat()
+
+        # add mandatory fields
+        newmessage[u'receivedtimestamp'] = toUTC(datetime.now()).isoformat()
+        newmessage[u'eventsource'] = u'squid'
+        newmessage[u'severity'] = u'INFO'
+        print(newmessage)
```
pwnbus: Remove debugging statement

mq/plugins/squidFixup.py
```diff
@@ -0,0 +1,144 @@
...
+        newmessage["details"] = {}
+
+        # move some fields that are expected at the event 'root' where they belong
+        if "HOST_FROM" in message:
+            newmessage["hostname"] = message["HOST_FROM"]
+        if "TAGS" in message:
+            newmessage["TAGS"] = message["tags"]
+        if "category" in message:
+            newmessage["category"] = message["category"]
+        newmessage[u"customendpoint"] = message["customendpoint"]
```
Phrozyn: Should line 94 and 95 also have if statements to be consistent with the rest of the code here?

Phrozyn: Also if these fields do not exist, what will be put in their place?

PR #1066, opened 2019-01-24
Create absolute path from path including ../

pwnbus:


pwnbus:
When writing unit tests, this makes it easier to show the entire path (with ../ expanded into the directory name).

Goes from
`$SRC/tests/alerts/../../alerts/lib/../deadman_generic.json`
to
`$SRC/alerts/deadman_generic.json`

PR #1072, opened 2019-01-28
Add generic env run

darakian:
This PR adds a make target which calls a copy of the default docker compose file which has an env file enabled for all services. This env file is user supplied via the main makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,8 @@ run-cloudy-mozdef: ## Run the MozDef containers necessary to run in AWS (`cloudy
 .PHONY: run-env-mozdef
 run-env-mozdef: ## Run the MozDef containers with a user specified env file. Run with make 'run-env-mozdef -e ENV=my.env'
 	$(shell test -f $(ENV) || touch $(ENV))
-	$(shell cp $(ENV) docker/compose/user_input.env)
-	docker-compose -f docker/compose/docker-compose-user-env.yml -p $(NAME) pull
-	docker-compose -f docker/compose/docker-compose-user-env.yml -p $(NAME) up -d
+	ENV_FILE=$(abspath $(ENV))
+	docker-compose -f docker/compose/docker-compose.yml -f docker/compose/docker-compose-user-env.yml -p $(NAME) up -d
 
 restart-cloudy-mozdef:
 	docker-compose -f docker/compose/docker-compose-cloudy-mozdef.yml -p $(NAME) restart
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,13 @@ run-cloudy-mozdef: ## Run the MozDef containers necessary to run in AWS (`cloudy
 	docker-compose -f docker/compose/docker-compose-cloudy-mozdef.yml -p $(NAME) pull
 	docker-compose -f docker/compose/docker-compose-cloudy-mozdef.yml -p $(NAME) up -d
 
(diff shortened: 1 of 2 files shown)


docker/compose/docker-compose-user-env.yml
```diff
@@ -0,0 +1,319 @@
+---
+version: '3.7'
+services:
+  nginx:
+    image: mozdef/mozdef_nginx
```
pwnbus: Instead of copy/pasting the entire docker-compose.yml file, we can use "inheritance" (docker overrides) here, so that we only have to redefine the env_file location. This could look like:
```
---
version: '3.7'
services:
  nginx:
    env_file:
      - $ENV_FILE
  mongodb:
    env_file:
      - $ENV_FILE
  kibana:
    env_file:
      - $ENV_FILE
  elasticsearch:
    env_file:
      - $ENV_FILE

etc
```

Makefile
```diff
@@ -31,6 +31,13 @@ run-cloudy-mozdef: ## Run the MozDef containers necessary to run in AWS (`cloudy
 	docker-compose -f docker/compose/docker-compose-cloudy-mozdef.yml -p $(NAME) pull
 	docker-compose -f docker/compose/docker-compose-cloudy-mozdef.yml -p $(NAME) up -d
 
+.PHONY: run-env-mozdef
+run-env-mozdef: ## Run the MozDef containers with a user specified env file. Run with make 'run-env-mozdef -e ENV=my.env'
+	$(shell test -f $(ENV) || touch $(ENV))
```
pwnbus: I'm not a big fan of writing a file to location within the git repo, so could we change this so the environment file would get referenced directly? This way, we could store the environment file in a location outside of the mozdef project directory. It could look something like (with the addition of using a docker override):

```
.PHONY: run-env-mozdef
run-env-mozdef: ## Run the MozDef containers with a user specified env file (relative or absolute paths). Run with make 'run-env-mozdef -e ENV=my.env'
	ENV_FILE=$(abspath $(ENV)) docker-compose -f docker/compose/docker-compose.yml -f docker/compose/docker-compose-user-env.yml -p $(NAME) up -d
```

docker/compose/docker-compose-user-env.yml
darakian: Ah indeed that is nicer. I wasn't aware that I could inherit with a docker compose file. Give me a bit and I'll update the file.

Makefile
darakian: How does the compose.yml file know about ENV_FILE in this case? Is it just known as a global environmental variable?

darakian: Gotcha. My branch has been updated. I'm unsure if the make file is correct. I followed your edits but broke up the variable assignment and the docker call. Also I'm unsure if the docker compose pull is required.

darakian:
In testing I'm getting errors with this option
```
[root@ip-172-23-4-100 MozDef]# make run-env-mozdef -e ENV=my.env
/root/MozDef/my.env
docker-compose -f docker/compose/docker-compose.yml -f docker/compose/docker-compose-user-env.yml -p mozdef up -d
WARNING: The ENV_FILE variable is not set. Defaulting to a blank string.
ERROR: /root/MozDef/docker/compose is not a file.
make: *** [run-env-mozdef] Error 1
```

PR #1076, opened 2019-01-30
Fix API calls to use next_offset as this is the newer way to paginate Note: only authentication logs support this

gdestuynder:
This *should* fix the duplicates. Needs a bit of testing.

See also https://community.duo.com/t/duo-admin-api-admin-v2-logs-authentication/3836/11

NOTE: I used black for formatting hence the formatting changes :| On the flip side this make our formatting proper ;-)

I added inline comments to the PR so that you can see exactly what changed and why

### DIFF
--- a/cron/duo_logpull.py
+++ b/cron/duo_logpull.py
@@ -6,8 +6,10 @@
 
 import sys
 from datetime import datetime, timedelta, tzinfo
+
 try:
     from datetime import timezone
+
     utc = timezone.utc
 except ImportError:
     # Hi there python2 user
@@ -20,6 +22,7 @@ except ImportError:
 
         def dst(self, dt):
             return timedelta(0)
+
     utc = UTC()
 from configlib import getConfig, OptionParser
 import json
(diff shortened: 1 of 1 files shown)


cron/duo_logpull.py
```diff
@@ -129,55 +135,73 @@ def process_events(mozmsg, duo_events, etype, state):
...
+    # same code with `next_offset` as authentication uses.
+    state = process_events(
+        mozmsg, duo.get_administrator_log(mintime=state["administration"] + 1), "administration", state
+    )
+    state = process_events(
+        mozmsg,
+        duo.get_authentication_log(
+            api_version=2,
+            limit="1000",
+            sort="ts:asc",
```
gdestuynder: this `sort` is the bug that took forever to figure out.

cron/duo_logpull.py
```diff
@@ -129,55 +135,73 @@ def process_events(mozmsg, duo_events, etype, state):
...
 def main():
     try:
-        state = pickle.load(open(options.statepath, 'rb'))
+        state = pickle.load(open(options.statepath, "rb"))
     except IOError:
         # Oh, you're new.
         # Note API v2 expect full, correct and within range timestamps in millisec so we start recently
         # API v1 uses normal timestamps in seconds instead
-        state = {'administration': 0, 'authentication': 1547000000000, 'telephony': 0}
+        state = {
```
gdestuynder: Added default offsets for all categories - only authentication really supports it, but this avoid running into compat issues with the file

cron/duo_logpull.py
```diff
@@ -69,18 +72,21 @@ def process_events(mozmsg, duo_events, etype, state):
...
+    elif etype == "authentication":
+        noconsume = ["timestamp", "host", "eventtype"]
     else:
         return
 
     # Care for API v2
-    if isinstance(duo_events, dict) and 'authlogs' in duo_events.keys():
-        duo_events = duo_events['authlogs']
+    if isinstance(duo_events, dict) and "authlogs" in duo_events.keys():
+        offset = duo_events["metadata"]["next_offset"]
```
gdestuynder: this is the new offset logic so that we save the offset sent by duo when we're done processing this batch of logs, only authentication supports this right now

PR #1078, opened 2019-01-31
Eval to set ENV_FILE

darakian:
This PR actually sets the variable `ENV_FILE`. It's only a partial solution though as the docker compose call is unaware of the variable. My make knowledge is lacking in finding a solution for that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ run-cloudy-mozdef: ## Run the MozDef containers necessary to run in AWS (`cloudy
 .PHONY: run-env-mozdef
 run-env-mozdef: ## Run the MozDef containers with a user specified env file. Run with make 'run-env-mozdef -e ENV=my.env'
 	$(shell test -f $(ENV) || touch $(ENV))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ run-cloudy-mozdef: ## Run the MozDef containers necessary to run in AWS (`cloudy
 .PHONY: run-env-mozdef
 run-env-mozdef: ## Run the MozDef containers with a user specified env file. Run with make 'run-env-mozdef -e ENV=my.env'
 	$(shell test -f $(ENV) || touch $(ENV))
-	ENV_FILE=$(abspath $(ENV))
+	$(eval ENV_FILE:=$(abspath $(ENV)))
```
pwnbus: Oh, right, yes. I think in what I was testing with, I had the ENV_FILE setting on the same line as the docker-compose line, so that docker-compose would run with that environment variable set, this also works though!

darakian: Curious. On my system I can't get the docker-compose line to respect the variable ENV_FILE. What did your single line solution look like?

PR #1091, opened 2019-02-04
Assert expected_alert is None in NegativeAlertTestCase. 

darakian:
Unless I misunderstand the NegativeAlertTestCase there should never be a case where an alert is created. Having the negative case take an expected alert is pretty counter intuitive when no alert should ever be created.

### DIFF
--- a/tests/alerts/negative_alert_test_case.py
+++ b/tests/alerts/negative_alert_test_case.py
@@ -2,4 +2,5 @@ from alert_test_case import AlertTestCase
 
 
 class NegativeAlertTestCase(AlertTestCase):
+    assert self.expected_alert is None, "NegativeAlertTestCase is used to test cases where an alert does not occur. Therefore expected_alert should be None."
     expected_test_result = False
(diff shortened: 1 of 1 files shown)


darakian:
@pwnbus Any chance I can get some help on namespacing here?

pwnbus:
@darakian We have https://github.com/mozilla/MozDef/blob/master/tests/alerts/alert_test_suite.py#L115 already in place that should do this logic, but the assert error text could probably be updated to be specific as to what we're checking for.

darakian:
So in my case I was able to create Negative tests which would happily accept an expected alert. Should this logic instead live within the alert test suite with a `if pos/neg then else` sort of structure?

pwnbus:
@darakian Huh, interesting. On my local dev environment, I just added an expected_alert to a test, and it failed accordingly. Do you have a working example on your end where that isn't happening?

darakian:
@pwnbus  I do not. I thought I had one on a branch but I can't find it and trying to recreate it I also get errors. I must have been working doing something else. Thanks for checking.

PR #1093, opened 2019-02-05
Update slack message format for alert

pwnbus:


pwnbus:
Makes the messages look something like: screen shot 2019-02-05 at 2 22 12 pm

PR #1102, opened 2019-02-14
Update rabbitmq version to 3.7

pwnbus:


Phrozyn:
Fails on: No package rabbitmq-server-3.7.11 available. This would need to be pulled from rabbitmq specific repos and not EPEL.

pwnbus:
@Phrozyn I updated the PR to use the preferred new rabbitmq repo so it should be all set to rereview!

PR #1103, opened 2019-02-14
Fix duplicate alert generic loader

pwnbus:
This fixes the problem where multiple generic alerts could not alert on the same event. Since we are modifying and updating the event (with the alerts that triggered as a result of it), we need to make sure we "refresh" after this action.


pwnbus:
This will fail as a result of the mozdef_util changes, but once those are initially approved, I'll push out a new version of mozdef_util which will fix the build failures.

PR #1106, opened 2019-02-15
Adding open and close abstraction for managing indexes

Phrozyn:


cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+import logging
+from datetime import datetime, date, timedelta
+from configlib import getConfig, OptionParser
+from logging.handlers import SysLogHandler
+
+from mozdef_util.utilities.toUTC import toUTC
+from mozdef_util.elasticsearch_client import ElasticsearchClient
+
+
+logger = logging.getLogger(sys.argv[0])
```
pwnbus: We can use the mozdef_util version of logger (https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/logger.py)

like `from mozdef_util.utilities.logger import logger`

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+logger.level=logging.WARNING
+formatter = logging.Formatter('%(asctime)s %(name)s %(levelname)s %(message)s')
+
+
+def esCloseIndices():
+    if options.output == 'syslog':
```
pwnbus: We can remove all of this logic (line 30 - 35) when we import the logger.

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+        sh = logging.StreamHandler(sys.stderr)
+        sh.setFormatter(formatter)
+        logger.addHandler(sh)
+    index_to_close = ''
+    logger.debug('started')
+    try:
+        es = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
+
+        indices = es.get_indices()
+        print(indices)
```
pwnbus: Do we still need this line?

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+        # calc dates for use in index names events-YYYYMMDD, alerts-YYYYMM, etc.
+        odate_month = date.strftime(toUTC(datetime.now()) - timedelta(days=int(options.index_age)), '%Y%m')
+        # examine each index in the .conf file
```
pwnbus: This comment isn't correct, because each index is pulled from elasticsearch, not a .conf file.

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+        # for rotation settings
+        for index in indices:
+            if index.find('events') == -1:
```
pwnbus: We can actually do `if 'events' in index` which I think is a bit more clear as to what we're trying to accomplish here.

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+                print "Index: %s will not be closed." % (index)
```
pwnbus: Instead of printing, can we use the logger (`logger.debug('sometext')`)?

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+            else:
+                odate = str(odate_month)
```
pwnbus: We don't have to cast this to a string, as it's already one when it's initially set.

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+                if index.find(odate) == -1:
```
pwnbus: Same comment as previous, we can do `if odate in index`

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+def esCloseIndices():
+    if options.output == 'syslog':
+        logger.addHandler(SysLogHandler(address=(options.sysloghostname, options.syslogport)))
+    else:
+        sh = logging.StreamHandler(sys.stderr)
+        sh.setFormatter(formatter)
+        logger.addHandler(sh)
+    index_to_close = ''
+    logger.debug('started')
+    try:
```
pwnbus: I think we want to move this `try except` block to where it's only around the specific logic to close the index. Right now, if we receive an ES connection error, we'll get an exception like `Unhandled exception while closing '', terminating`

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+        # for rotation settings
+        for index in indices:
+            if index.find('events') == -1:
+                print "Index: %s will not be closed." % (index)
+            else:
+                odate = str(odate_month)
+                if index.find(odate) == -1:
+                    print "Index: %s doesn't meet aging requirements and will not be closed" % (index)
+                else:
+                    index_to_close = index
```
pwnbus: Instead of creating this `index_to_close` variable, can we just use the `index` variable instead? This way, we also don't have to do `index_to_close = ''` earlier up the file.

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+                    print "Index: %s will be closed." % (index_to_close)
```
pwnbus: Can we use the logger here?

cron/closeIndices.py
```diff
@@ -0,0 +1,101 @@
...
+                    es.index_close(index_to_close)
+                    options.index_age = int(options.index_age) + 1
```
pwnbus: I don't like how we're modifying the options.index_age config value here. I think this is due to the comparison logic to determine if an index is old enough to be closed.

I think if we instead took the the index name from ES, converted that string `events-201904` to a date, and then compared that date to date(options.index_age) and if it was older, we call the close index function.

mozdef_util/mozdef_util/elasticsearch_client.py
```diff
@@ -38,6 +38,12 @@ def __init__(self, servers, bulk_amount=100, bulk_refresh_time=30):
         self.es_connection.ping()
         self.bulk_queue = BulkQueue(self, threshold=bulk_amount, flush_time=bulk_refresh_time)
 
+    def close_index(self, index_name)
```
pwnbus: This is a syntax error, you need a semicolon at the end of the line.

mozdef_util/mozdef_util/elasticsearch_client.py
```diff
@@ -38,6 +38,12 @@ def __init__(self, servers, bulk_amount=100, bulk_refresh_time=30):
...
+        self.es_connection.indices.close(index=index_name)
+
+    def open_index(self, index_name)
```
pwnbus: Syntax error, missing a trailing semi colon.

tests/mozdef_util/test_elasticsearch_client.py
```diff
@@ -120,6 +120,27 @@ def test_search_no_results(self):
         assert results['hits'] == []
 
 
+class TestCloseIndex(ElasticsearchClientTest):
+
+    def test_index_close(self):
+        if pytest.config.option.delete_indexes:
+            self.es_client.create_index('test_index')
+        time.sleep(1)
+        indices = self.es_client.index_close('test_index')
```
pwnbus: I think these tests can be improved to verify more of the open/close logic. All we are doing here is checking that we can run 2 functions, and then that the second function, returns True.

For example, what if we closed the events index, and then tried to save an event. When you try writing to a closed index, elasticsearch throws an exception, so we can tell pytest to expect an exception like https://github.com/mozilla/MozDef/blob/master/tests/mozdef_util/query_models/test_search_query.py#L291

tests/mozdef_util/test_elasticsearch_client.py
```diff
@@ -120,6 +120,27 @@ def test_search_no_results(self):
...
+        assert indices is True
+
+
+class TestOpenIndex(ElasticsearchClientTest):
+
+    def test_index_open(self):
```
pwnbus: We can improve these tests by closing an index, opening the index, and verifying that we can write an event.

cron/closeIndices.py
Phrozyn: ack will revise.

cron/closeIndices.py
Phrozyn: Will do

cron/closeIndices.py
Phrozyn: Thanks, was part of debugging :)

cron/closeIndices.py
Phrozyn: Edited, leftover artifacts from the maint cron.

cron/closeIndices.py
Phrozyn: I feel like it's evident, but I'm ok with changing this to what you feel is more appropriate.

cron/closeIndices.py
Phrozyn: I like your approach, sounds cleaner. Will modify

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: The first test closes the test index tha L#127 creates.
The next test creates a test index on L#137 and then closes it on L#139, and then opens the one we closed on L#140.

Are you just asking for a negative test case for saving to a closed index? That will never happen in production as we only save to the current events alias, which is never closed. We only close  "actual index names" not aliases. In fact in 6.x you can't close an alias at all, you MUST cite the index name.  We should never, ever, be writing to any indexes that are not tagged with the current events alias.

Please help me find the value in this request or explain how it will benefit us moving forward.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: My previous comment stands, we should never write an event to any index that is not the current "events" tagged index. No events tagged index should ever be closed. This is not a use case we would ever pursue.

mozdef_util/mozdef_util/elasticsearch_client.py
Phrozyn: fixed in last commit.

mozdef_util/mozdef_util/elasticsearch_client.py
Phrozyn: fixed in last commit.

cron/closeIndices.py
Phrozyn: still need to do

Phrozyn: To explain, it says don't close any indexes unless they are 30 days or older. That's how that logic is working.

cron/closeIndices.py
Phrozyn: So I should have two tries?
One that gives us an exception for when it cannot connect, and one for when it can't connect to close?

cron/closeIndices.py
Phrozyn: sure! Thanks for the suggestion :)

cron/closeIndices.py
Phrozyn: when I was testing it required it, because odate_month is actually turned into an int later on in order to increment it to 30+ days

tests/mozdef_util/test_elasticsearch_client.py
pwnbus: What I'm saying is, you aren't actually testing the functionality that an index does get closed, you're testing that `self.es_client.index_close('test_index')` returns a True value.

For example, if this is what the implementation looks like of index_close, your tests will pass, which is not what we want:
```
def index_close(index_name):
    return True
```

So, to improve these tests, we want to have the tests verify that `self.es_client.index_close('test_index')` actually closes an index (that that function is doing what we want it to do). 

One way to verify if an index actually gets closed, is to try and write to it and expecting an elasticsearch closed index write error to get thrown.

Another idea might be to query elasticsearch and have them tell us if the `test_index` is now closed.

cron/closeIndices.py
Phrozyn: What this does is say if any of the indexes do not have "events" in the name, then they won't be closed. Basically, any indexes that are not named "events-something" are never rotated, and also never closed.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: I understand now, thank you for clarifying.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: Agreed!

cron/closeIndices.py
Phrozyn: Done :)

cron/closeIndices.py
Phrozyn: Take a look at what I added and let me know if that works for you.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: Done!

cron/closeIndices.py
```diff
@@ -0,0 +1,91 @@
...
+            index_date = index.rsplit('-', 1)[1]
+            logger.debug("Checking to see if Index: %s can be closed." % (index))
+            month_ago_date = toUTC(datetime.now()) - timedelta(days=options.index_age)
+            point_of_close = month_ago_date.strftime('%Y%m%d')
+            if len(index_date) == 8:
+                month_ago_str = point_of_close
+                index_date_obj = datetime.strptime(index_date, '%Y%m%d')
+                index_date_str = index_date_obj.strftime('%Y%m%d')
+                try:
+                    if int(month_ago_str) > int(index_date_str):
```
pwnbus: The date objects already implement a comparison function, so we don't need to convert these to int, we can do simply `if month_ago_date > index_date_obj:`

You may need to remove the timezone from the month_ago_date before the if statement, but that can be done by `month_ago_date = month_ago_date.replace(tzinfo=None)`

cron/closeIndices.py
```diff
@@ -0,0 +1,91 @@
...
+    except Exception as e:
+        logger.error("Unhandled exception while connecting to ES, terminating: %r" % (e))
+
+    # examine each index pulled from get_indice
+    # to determine if it meets aging criteria
+    for index in indices:
+        if 'events' in index:
+            index_date = index.rsplit('-', 1)[1]
+            logger.debug("Checking to see if Index: %s can be closed." % (index))
+            month_ago_date = toUTC(datetime.now()) - timedelta(days=options.index_age)
```
pwnbus: You need to convert options.index_age to an integer, or else this line stack traces.

tests/mozdef_util/test_elasticsearch_client.py
```diff
@@ -120,6 +120,54 @@ def test_search_no_results(self):
...
 
 
+class TestCloseIndex(ElasticsearchClientTest):
+
+    def teardown(self):
+        super(TestCloseIndex, self).teardown()
+        if pytest.config.option.delete_indexes:
+            self.es_client.delete_index('test_index')
+
+    def test_writing_to_closed_index(self):
```
pwnbus: This function name is misleading, since we aren't "writing" anything.

cron/closeIndices.py
pwnbus: We can also move this line to right before the for loop (line 32), since we only have to set this variable once.

cron/closeIndices.py
Phrozyn: Thanks @pwnbus I was not having success in understanding how to do this efficiently.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: I don't think it's misleading, it's exactly what the test is attempting to do. 
Maybe adding what you feel is more appropriate would be more helpful. 
Do you want to see something like test_failure_writing_to_closed_index?

cron/closeIndices.py
Phrozyn: Good call, though in my tests it behaves as an integer, so not sure what changes between my environment and yours.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: Oh nvm! I see what happened :) good catch!

Phrozyn:
This was tested to work in my dev environment, all requested changes have been implemented.

CHANGELOG
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+## [v1.38.4] - 2019-04-07
+### Added
+- Elasticsearch Client abstraction API to mozdef-util for opening or closing indices.
```
pwnbus: We do not need to modify this file right now, this is only when we do a mozdef release.

We only need to modify https://github.com/mozilla/MozDef/blob/master/mozdef_util/HISTORY.rst when we're doing a mozdef-util release, which you have done correctly.

PR #1112, opened 2019-02-26
Adjust proxy alerts

mpurzynski:


tests/alerts/test_proxy_drop_executable.py
```diff
@@ -31,36 +30,39 @@ class TestAlertProxyDropExecutable(AlertTestSuite):
     # This event is the default negative event that will not cause the
     # alert to trigger
     default_negative_event = AlertTestSuite.copy(default_event)
-    default_negative_event["_source"]["details"]["destination"] = "http://foo.mozilla.com/index.html"
+    default_negative_event["_source"]["details"][
```
pwnbus: I think this change actually makes it more difficult to read. Since I'm assuming these changes were automatically done via some tool, can you configure that tool to ignore line length edits like these in the future please?

mpurzynski: https://github.com/ambv/black is what I started using for everything. Do you want me to stop splitting lines and rather keep them long? I guess I can configure that https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines

PR #1114, opened 2019-02-27
[WIP] Remove doc_type from cron

Phrozyn:
**Completed:**
- Added logger from mozdef_util and stripped out old logging function with these edits where applicable.
- Update to default template
- Update to rest/index.py for dashboards.
- **healthandstatus** removed doc_type in favor of default elasticsearch_client setting this.
  - Update to defaultMozdefState index mapping
- **healthtomongo**
  - set _type to be "type" = "mozdefhealth"
- **correlateUserMacAddress** sets doc_type to 'usernamemacaddress', change to "_doc"
- **okta2mozdef** removed doc_type in favor of default elasticsearch_client setting this.
- **sqs_queue_status** removed doc_type in favor of default elasticsearch_client setting this.
- **verifyeventfields**
  - removed doc_type in favor of default elasticsearch_client setting this.
  - determine if the check for _type is still needed, or if we want to change it to "type"
    - changed to type for now, can always revert to _type, but with how it excludes mozdefhealth and stats, which are now moved to "type", I felt it was best to preserve the current intended functionality.

**Work To Do:**

- **ESWorkers:**
  - move any custom doc_type being set in any esworker to be a part of the parent structure with the designation of  "type"
  - change actual doc_type in all esworkers workers to be "_doc"
  - **MQ Plugins**
    - Various plugins set doc_type, must be changed to set "type"

- **Alerts:**
  - move any custom doc_types to "type" within alerts.
    - change actual doc_type in all alerts to be "_doc"

- **Cron**
  - **collectattackers**
    - comments say it sets _type, but there was no code that directly sets this that I could find.

- **Elasticsearch Client**
  - **elasticsearch_client.py:** L197: change doc_type to "doc"

- **TESTS**
  - **test_promisc_audit.py**
    - set _type to event or look for "type = auditd" instead
  - **test_write_audit.py**
    - set _type to event or look for "type = auditd" instead
  - **test_cloudtrail_logging_disabled**
    - set _type to event or look for "type = cloudtrail" instead

- **DOCKER**
  - **initial_setup.py**
    - dashboard is no longer "_type", but "type"
    - visualization is no longer "_type", but "type"

- **EXTRA**
  - modify whatever sets empty fields to UNKNOWN to ensure that it no longer sets _type as "UNKNOWN"
  - **Benchmarking**
    - insert_bulk.js needs change _type to "_doc" from "blog"

- **Kibana:**
  - Determine if kibana index name should be cast into a variable
    - This index name appears to get created with a random number (e.g: .kibana_1, kibana_7)


cron/vidyo2MozDef.py
```diff
@@ -42,18 +42,22 @@ class MozDefEvent():
     log['severity'] = 'INFO'
     log['summary'] = None
     log['category'] = 'event'
+    log['source'] = 'vidyo'
     log['tags'] = list()
     log['details'] = dict()
 
-    def __init__(self, url='http://localhost/events', summary=None, category='event', severity='INFO', tags=[], details={}):
+
+    def __init__(self, url='http://localhost/events', summary=None, category='event', severity='INFO', source='vidyo', tags=[], details={}):
```
Phrozyn: Added source here, because I was not sure if "type" could be used when stating self.type = type in that fashion.
I felt maybe we could go  back and change it to add casting this value into a type field, but for now added it as source. pinging pwnbus for insight.

cron/sqs_queue_status.py
```diff
@@ -100,11 +100,11 @@ def getQueueSizes():
         qcounter -= 1
     # post to elasticsearch servers directly without going through
     # message queues in case there is an availability issue
-    es.save_event(index=options.index, doc_type='mozdefhealth', body=json.dumps(healthlog))
+    es.save_event(index=options.index, body=json.dumps(healthlog))
```
pwnbus: We need to modify the event before these lines to set the type = 'mozdefhealth'

cron/syncAlertsToMongo.py
```diff
@@ -6,47 +6,25 @@
...
-
 
 def genMeteorID():
     return('%024x' % random.randrange(16**24))
 
 
 def getESAlerts(es):
     search_query = SearchQuery(minutes=50)
-    search_query.add_must(TermMatch('_type', 'alert'))
+    search_query.add_must(TermMatch('type', 'alert'))
```
pwnbus: In https://github.com/mozilla/MozDef/pull/1129 we aren't setting the type here to alert anymore, so this logic will then be broken.

cron/syncAlertsToMongo.py
```diff
@@ -74,15 +52,15 @@ def updateMongo(mozdefdb, esAlerts):
...
+            # capture the elastic search meta data (index/id/type)
             # set the date back to a datetime from unicode, so mongo/meteor can properly sort, select.
             mrecord['utctimestamp']=toUTC(mrecord['utctimestamp'])
             # also set an epoch time field so minimongo can sort
             mrecord['utcepoch'] = calendar.timegm(mrecord['utctimestamp'].utctimetuple())
             mrecord['esmetadata'] = dict()
             mrecord['esmetadata']['id'] = a['_id']
             mrecord['esmetadata']['index'] = a['_index']
-            mrecord['esmetadata']['type'] = a['_type']
+            mrecord['esmetadata']['type'] = a['type']
```
pwnbus: This isn't the correct location of `type`, it's in `a['_source']['type']

pwnbus:
I reverted any changes that was specific to ES6 breaking changes, and broke those out into a separate PR. This PR is now solely for doc_type removal changes.

cron/sqs_queue_status.py
pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1114/commits/bf3a143b0c9b5058fcc6221d776e2724a2d5d7ac

cron/syncAlertsToMongo.py
pwnbus: I think we can actually delete this line entirely, as the type of the event is no longer esmetadata.

PR #1118, opened 2019-03-05
Resolving issues with sshd events not parsing correctly.

Phrozyn:
Strips leading whitespace from summary field.
Parses sshd events based on processname which is built by MQ worker first, as I found that not all events have a details.program. Rather than duplicate fields of the same data, I determined it was better to use the field present across all events.

### DIFF
--- a/mq/plugins/parse_sshd.py
+++ b/mq/plugins/parse_sshd.py
@@ -10,8 +10,8 @@ class message(object):
 
     def __init__(self):
         '''
-        takes an incoming sshd message
-        and sets the doc_type
+        takes an incoming sshd message and
+        parses sshd details into fields
         '''
 
         self.registration = ['sshd']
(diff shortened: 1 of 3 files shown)


mq/plugins/parse_sshd.py
```diff
@@ -23,35 +23,51 @@ def onMessage(self, message, metadata):
         self.session_opened_regex = re.compile('^pam_unix\(sshd\:session\)\: session (opened|closed) for user (?P<username>[a-zA-Z0-9\@._-]+)(?: by \(uid\=\d*\))?$')
         self.postponed_regex = re.compile('^Postponed (?P<authmethod>\w+) for (?P<username>[a-zA-Z0-9\@._-]+) from (?P<sourceipaddress>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}) port (?P<sourceport>\d{1,5}) ssh2(?: \[preauth\])?$')
         self.starting_session_regex = re.compile('^Starting session: (?P<sessiontype>\w+)(?: on )?(?P<device>pts/0)? for (?P<username>[a-zA-Z0-9\@._-]+) from (?P<sourceipaddress>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}) port (?P<sourceport>\d{1,5})$')
+        self.unauthorized_user_regex = re.compile('^Invalid user (?P<username>[a-zA-Z0-9\@._-]+) from (?P<sourceipaddress>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})?$')
+        self.userauth_request_regex = re.compile('^input_userauth_request: invalid user (?P<username>[a-zA-Z0-9\@._-]+) \[preauth\]')
+        self.disconnect_regex = re.compile('^Received disconnect from (?P<sourceipaddress>\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}): (?P<sourceport>\d{1,5}): (|Bye Bye|Normal Shutdown, Thank you for playing) \[preauth\]')
```
pwnbus: Do we have some sample events that match these regex, so we can add some additional tests to https://github.com/mozilla/MozDef/blob/master/tests/mq/plugins/test_parse_sshd.py?

Phrozyn: Added! Let me know if this is enough or if you think we need more.

PR #1126, opened 2019-03-18
Update mozdef-util for ES6 upgrade

pwnbus:
This will require creating a new mozdef-util package when we're ready.


pwnbus:
We decided to break up the doc_type removal work from the ES6 breaking changes, so this PR will get expanded out into others, but so far, https://github.com/mozilla/MozDef/pull/1179 https://github.com/mozilla/MozDef/pull/1180 and https://github.com/mozilla/MozDef/pull/1181

PR #1129, opened 2019-03-19
[WIP] Alert changes (doc_type removal)

Phrozyn:
This removes the doc_type metadata entries as the elasticsearch client will provide that data.

### DIFF
--- a/alerts/lib/alerttask.py
+++ b/alerts/lib/alerttask.py
@@ -393,7 +393,6 @@ class AlertTask(Task):
         for e in events:
             alert['events'].append({
                 'documentindex': e['_index'],
-                'documenttype': e['type'],
                 'documentsource': e['_source'],
                 'documentid': e['_id']})
         self.log.debug(alert)
@@ -445,7 +444,6 @@ class AlertTask(Task):
                     event['_source']['alerts'] = []
                 event['_source']['alerts'].append({
                     'index': alertResultES['_index'],
-                    'type': alertResultES['type'],
                     'id': alertResultES['_id']})
 
                 if 'alert_names' not in event['_source']:
--- a/alerts/lib/alerttask.py
+++ b/alerts/lib/alerttask.py
@@ -393,7 +393,7 @@ class AlertTask(Task):
         for e in events:
             alert['events'].append({
                 'documentindex': e['_index'],
-                'documenttype': e['_type'],
+                'documenttype': e['type'],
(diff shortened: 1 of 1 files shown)


jeffbryner:
Can I ask why this instead of removing _type? (since as I understand it, it will always be static)? does type==category?

Phrozyn:
> Can I ask why this instead of removing _type? (since as I understand it, it will always be static)? does type==category?

Since we have different _type's of alerts, this will now be cast into "type" rather than "_type" since "_type" will become _doc and is no longer going to be used for filtering. We felt if we preserved the _type as type we'd still be able to use it for filtering. It's a sub-category essentially.

jeffbryner:
_type currently for alerts is always 'alert' and I don't see a place where that's different? 
https://github.com/mozilla/MozDef/blob/4190c8d5c5fb79485f56d6cca86e0227daf5336e/alerts/lib/alerttask.py#L377

alerts/lib/alerttask.py
```diff
@@ -393,7 +393,7 @@ def createAlertDict(self, summary, category, tags, events, severity='NOTICE', ur
         for e in events:
             alert['events'].append({
                 'documentindex': e['_index'],
-                'documenttype': e['_type'],
+                'documenttype': e['type'],
```
Phrozyn: in my head I feel like maybe I should remove documenttype from this stanza, and add t his later as simply 'type' so that there is no confusion between _type: _doc and type: <alert type>

Phrozyn:
> _type currently for alerts is always 'alert' and I don't see a place where that's different?
> 
> [MozDef/alerts/lib/alerttask.py](https://github.com/mozilla/MozDef/blob/4190c8d5c5fb79485f56d6cca86e0227daf5336e/alerts/lib/alerttask.py#L377)
> 
> Line 377 in [4190c8d](/mozilla/MozDef/commit/4190c8d5c5fb79485f56d6cca86e0227daf5336e)
>  def createAlertDict(self, summary, category, tags, events, severity='NOTICE', url=None, ircchannel=None):

You're right, I just looked again (earlier I had seen auditd and alert) and there is only one _type.
I guess my session had timed out and it mixed results in a weird way. I'll modify.

pwnbus:
Closing PR since we've decided to change the destination branch to doc_type_removal. These changes have been cherry-picked into https://github.com/mozilla/MozDef/pull/1182

PR #1133, opened 2019-03-20
Add initial CHANGELOG (captures changes in v1.37)

tristanweir:


andrewkrug:
@tristanweir Thanks so much for doing this!  I'm looking forward to keeping up on the CHANGELOG in the future.

PR #1147, opened 2019-03-21
Ipaddr alert plugin

mpurzynski:


Phrozyn:
Hi Michal, I'm trying to understand the purpose of the ipaddr addition and how that differs from the work already done by the mq plugin ipFixup which extracts the IP information out of the event and assigns it accordingly. This looks as though it's duplicating that plugin, which should already have pulled this information out and structured the data which will be present in the alert.

A little more info around the workflow and the way this adds value would be awesome.

alerts/plugins/ipaddr.py
```diff
@@ -0,0 +1,132 @@
...
+            Elastic search is inconsistent about returning IPs as
+            decimal or IPs.
+            In a query an IP field is returned as string.
+            In a facets an IP field is returned as decimal.
+            No ES field type exists for ipv6, so always having
+            a string version is the most flexible option.
+        '''
+
+        # here is where you do something with the incoming alert message
+        if 'events' in message.keys():
```
pwnbus: We're trying to get away from calling `message.keys()`, we can instead do the same logic by `if 'events' in message:`

alerts/plugins/ipaddr.py
```diff
@@ -0,0 +1,132 @@
...
+    if isinstance(message['errors'], list):
+        message['errors'].append(error)
+
+
+class message(object):
+    def __init__(self):
+        '''
+        takes an incoming alert
+        and uses it to trigger an event using
+        the pager duty event api
```
pwnbus: Copy/pasta incorrect comments

mpurzynski:
The goal of this plugin is to surface the details.sourceipaddress or the
details.destinationipaddress field, if present in events that lead to an
alert.
We do not propagate it currently - the search for

_exists_:details.sourceipaddress

In the alerts index came up empty.

The IP address plugin for the MQ worker adds IP addresses to events, but it
won't help much for alerts, hence the alert plugin.

Phrozyn:
That makes sense, thanks for the explanation! Adding that should make finding alerts around a particular IP easier when searching Kibana or correlating multiple alerts.

alerts/plugins/iqrisk.py
```diff
@@ -0,0 +1,86 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import os
+from configlib import getConfig, OptionParser
+import sys
+from datetime import datetime, timedelta
+sys.path.append(os.path.join(os.path.dirname(__file__), "../lib"))
+from iqrlib import iqr
```
pwnbus: Is this a new library? if so, we need to add it to requirements.txt

mpurzynski: I removed this import as it was not needed here. The IRQ library will be in a separate PR.

alerts/plugins/ipaddr.py
mpurzynski: Comments corrected to reflect the reality.

alerts/plugins/ipaddr.py
mpurzynski: Changed all instances of .keys() to if "a" in b

PR #1164, opened 2019-03-22
Clean up PHONY lines in Makefile

gene1wood:


gene1wood:
@gdestuynder I want your feedback on if this makes sense. I've done a few things here with the PHONY lines, some which by my understanding are required and some which are just subjective cosmetic changes.

* [Removing the multiple PHONY definitions in one line which mirrors the target line calling it's child targets](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965L108). This seems confusing to me since these are likely set as PHONY later next to the child targets. Also there are [cases where the two don't match](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965L108) which caused me to make a mistake, read the PHONY line instead of the target line and misunderstand what was being called in what order
* [Put each tagets associated PHONY line on the line before the target](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965R48). Though this is cosmetic I think it will help prevent [mistakes where people miss adding a PHONY line](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965R65)
* I guess I should be consistent and either not [break up these "alias" targets and use a single PHONY line](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965R48) or I should [break them up](https://github.com/mozilla/MozDef/pull/1164/files#diff-b67911656ef5d18c4ae36cb6741b7965R82). Do you think it would be good to break these all up so there's always a single PHONY line for each target?

gdestuynder:
it seems right to me
i tend to try to put one phony per target so that i dont lose track of them so +1 on this change. i wish makefiles allowed to globally default to phony as more and more people use them like this instead of tracking dependencies for C files or object code

gene1wood:
Cool, I'll make another PR that does this in all cases

PR #1165, opened 2019-03-22
Fix missing packer container fetch and other fixes

### LINKED ISSUE #1163: march_swarm Makefile bug causing wrong images and duplicate builds
So the addition of `make BRANCH=${BRANCH} set-version-and-fetch-docker-container` to the `deploy` script in

https://github.com/mozilla/MozDef/pull/1141/files#diff-c5581172c41e86058fcfb70d4bc45329R32

is what causes AWS CodeBuild to build the docker containers and write them into the AMI

It does this by
* Following the commands in `buildspec.yml`
* Which say to run `bash cloudy_mozdef/ci/deploy` which
  * Calls `build-from-cwd` make target which runs docker-compose BUILD for docker/compose/docker-compose.yml (not docker-compose-cloudy-mozdef.yml)
  * Calls `hub-login` make target
  * Calls `packer-build-github` make target in cloudy mozdef makefile which calls `ci/pack_and_copy` which runs packer to build the AMI and replicate it using `packer.json`. Within the packer temp AMI
    * git clones the MozDef GitHub repo locally and checks out the correct BRANCH
    * Calls `set-version-and-fetch-docker-container`
      * which calls `build-from-cwd` make target which builds the containers a second time (again from the wrong compose file)
      * and calls `tag-images` make target which tags the built docker containers
      * and replaces `latest` with the BRANCH in docker/compose/docker-compose-cloudy-mozdef.yml so that when the instance launches and runs `run-cloudy-mozdef` it will have the right docker compose file
  * Calls `publish-versioned-templates` make target in cloudy mozdef makefile which calls `ci/publish_versioned_templates` which versions CloudFormation templates and uploads them to S3
  * Calls `set-version-and-fetch-docker-container` make target 
    * which calls `build-from-cwd` make target which builds the containers a third time (again from the wrong compose file)
    * and calls `tag-images` make target which tags the built docker containers
    * and replaces `latest` with the BRANCH in docker/compose/docker-compose-cloudy-mozdef.yml which isn't used here in CodeBuild
  * Calls `docker-push-tagged` which calls
    * `tag-images` which tags the docker images with the BRANCH a second time
    * `hub-tagged` which uploads the local tagged BRANCH images to dockerhub

1. The docker images built to be used by packer are for docker-compose.yml not for docker-compose-cloudy-mozdef.yml and so aren't putting the cloudy mozdef specific images (like mozilla.oidc.accessproxy) in the packer image
2. Should we move `hub-login` in `deploy` down to before `docker-push-tagged` since that's why we're logging in
3. By calling `set-version-and-fetch-docker-container` we are building the docker containers
   * First in codebuild
   * Next inside the packer temporary ec2 instance
   * Finally again in codebuild
4. By calling `set-version-and-fetch-docker-container` twice, once in CodeBuild and once in the packer ec2 instance we're tagging the images 3 times
5. Are the docker images supposed to be built in codebuild or on the packer ec2 instance?

gene1wood:
* Add docker-compose pull to packer.json to fetch locally packages other than those we build Fixes #1163 
* Remove CodeBuild call to set-version-and-fetch-docker-container make target
  This is because
  * We've already built the containers in CodeBuild and don't need to do it again
  * We don't need to tag the docker images because `docker-push-tagged` will trigger that
  * We don't need to modify docker/compose/docker-compose-cloudy-mozdef.yml because we don't use it in CodeBuild
* Add build make target that mirrors previous behavior
  * Previous behavior of build was the build-from-cwd behavior
* Make the sed regex more specific so it only affects MozDef container images

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ run-tests: run-tests-resources  ## Just run the tests (no build/get). Use `make 
 	docker run -it --rm --network=test-mozdef_default mozdef/mozdef_tester bash -c "source /opt/mozdef/envs/python/bin/activate && py.test --delete_indexes --delete_queues $(TEST_CASE)"
 rebuild-run-tests: build-tests run-tests
 
+.PHONY: build
+build: build-from-cwd
+
 .PHONY: build-from-cwd
 build-from-cwd:  ## Build local MozDef images (use make NO_CACHE=--no-cache build to disable caching)
 	docker-compose -f docker/compose/docker-compose.yml -p $(NAME) $(NO_CACHE) $(BUILD_MODE)
--- a/Makefile
+++ b/Makefile
@@ -153,4 +153,4 @@ new-alert: ## Create an example alert and working alert unit test
 
 .PHONY: set-version-and-fetch-docker-container
 set-version-and-fetch-docker-container: build-from-cwd tag-images # Lock the release of MozDef by pulling the docker containers on AMI build and caching replace all instances of latest in the compose override with the BRANCH
-	sed -i s/latest/$(BRANCH)/g docker/compose/docker-compose-cloudy-mozdef.yml
+	sed -i 's;\image: mozdef/\(.*\):latest$;image: mozdef/\1:$(BRANCH);g' docker/compose/docker-compose-cloudy-mozdef.yml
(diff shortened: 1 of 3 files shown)


gene1wood:
To test this we need toc confirm that
* the sed line is causing the compose file in the AMI to reference the branch name for the container tag
* the containers aren't built and tagged a second time in CodeBuild
* the ec2 instance on launch no longer pulls down mozilla.oidc.accessproxy from dockerhub

PR #1179, opened 2019-03-25
Update Elasticsearch6 docker container

pwnbus:


docker/compose/elasticsearch/files/elasticsearch.yml
```diff
@@ -1,98 +1,16 @@
...
 path.logs: /var/log/elasticsearch
 path.data: /var/lib/elasticsearch
 http.host: 0.0.0.0
 discovery.type: single-node
 action.destructive_requires_name: true
-action.auto_create_index: false
-# Mozdef Custom Settings #
+
+# Disable auto creation unless these indexes
+action.auto_create_index:  .watches,.triggered_watches,.watcher-history-*
```
Phrozyn: The kibana index should be added here. ".kibana*"

pwnbus: I don't think it needs to be, because we create the .kibana as part of the initial setup of the docker environment.

Phrozyn: Ahh yeah, this is docker :p duh, ok nevermind me :)

PR #1130, opened 2019-03-19
[WIP] MQ worker changes (doc_type/type)

Phrozyn:
modified:   mq/esworker_cloudtrail.py
        modified:   mq/esworker_eventtask.py
        modified:   mq/esworker_papertrail.py
        modified:   mq/esworker_sns_sqs.py
        modified:   mq/esworker_sqs.py
        modified:   mq/plugins/auditdFixup.py
        modified:   mq/plugins/broFixup.py
        modified:   mq/plugins/complianceitems.py
        modified:   mq/plugins/customDocType.py
        modified:   mq/plugins/dropMessage.py
        modified:   mq/plugins/googleFixup.py
        modified:   mq/plugins/ipFixup.py
        modified:   mq/plugins/netflowFixup.py
        modified:   mq/plugins/parse_sshd.py
        modified:   mq/plugins/parse_su.py
        modified:   mq/plugins/squidFixup.py
        modified:   mq/plugins/suricataFixup.py
        modified:   mq/plugins/vidyoCallID.py
        modified:   mq/plugins/vulnerability.py

- Removed doc_type (_type) references.
- Added type references (which take on the value that _type held).
- Modified verbiage in some files for accuracy.

Must inform about vulnerability.py changes to @arcrose prior to committing this change to master.

PR #1183, opened 2019-03-25
Remove doc_type from mq

pwnbus:
Replaces https://github.com/mozilla/MozDef/pull/1130

        modified:   mq/esworker_cloudtrail.py
        modified:   mq/esworker_eventtask.py
        modified:   mq/esworker_papertrail.py
        modified:   mq/esworker_sns_sqs.py
        modified:   mq/esworker_sqs.py
        modified:   mq/plugins/auditdFixup.py
        modified:   mq/plugins/broFixup.py
        modified:   mq/plugins/complianceitems.py
        modified:   mq/plugins/customDocType.py
        modified:   mq/plugins/dropMessage.py
        modified:   mq/plugins/googleFixup.py
        modified:   mq/plugins/ipFixup.py
        modified:   mq/plugins/netflowFixup.py
        modified:   mq/plugins/parse_sshd.py
        modified:   mq/plugins/parse_su.py
        modified:   mq/plugins/squidFixup.py
        modified:   mq/plugins/suricataFixup.py
        modified:   mq/plugins/vidyoCallID.py
        modified:   mq/plugins/vulnerability.py

- Removed doc_type (_type) references.
- Added type references (which take on the value that _type held).
- Modified verbiage in some files for accuracy.

Must inform about vulnerability.py changes to @arcrose prior to committing this change to master.

### DIFF
--- a/mq/esworker_cloudtrail.py
+++ b/mq/esworker_cloudtrail.py
@@ -259,6 +259,7 @@ def keyMapping(aDict):
             returndict['utctimestamp'] = toUTC(datetime.now()).isoformat()
         if 'type' not in returndict:
             # default replacement for old _type subcategory.
+            # to preserve filtering capabilities
             returndict['type'] = 'cloudtrail'
     except Exception as e:
         logger.exception(e)
--- a/mq/esworker_cloudtrail.py
+++ b/mq/esworker_cloudtrail.py
@@ -221,6 +221,7 @@ def keyMapping(aDict):
 
             elif k in ('type', 'eventtype', 'category'):
                 returndict[u'category'] = toUnicode(v)
+                returndict[u'type'] = 'cloudtrail'
 
             # custom fields as a list/array
             elif k in ('fields', 'details'):
@@ -257,10 +258,7 @@ def keyMapping(aDict):
         if 'utctimestamp' not in returndict:
             # default in case we don't find a reasonable timestamp
             returndict['utctimestamp'] = toUTC(datetime.now()).isoformat()
-        if 'type' not in returndict:
-            # default replacement for old _type subcategory.
-            # to preserve filtering capabilities
-            returndict['type'] = 'cloudtrail'
+
     except Exception as e:
         logger.exception(e)
         logger.error('Malformed message: %r' % aDict)
--- a/mq/esworker_cloudtrail.py
+++ b/mq/esworker_cloudtrail.py
@@ -257,6 +257,9 @@ def keyMapping(aDict):
         if 'utctimestamp' not in returndict:
             # default in case we don't find a reasonable timestamp
             returndict['utctimestamp'] = toUTC(datetime.now()).isoformat()
+        if 'type' not in returndict:
+            # default replacement for old _type subcategory.
+            returndict['type'] = 'cloudtrail'
     except Exception as e:
         logger.exception(e)
         logger.error('Malformed message: %r' % aDict)
@@ -369,7 +372,6 @@ class taskConsumer(object):
             # default elastic search metadata for an event
             metadata = {
                 'index': 'events',
-                'doc_type': 'cloudtrail',
                 'id': None
             }
             # just to be safe..check what we were sent.
@@ -416,7 +418,6 @@ class taskConsumer(object):
                 self.esConnection.save_event(
                     index=metadata['index'],
                     doc_id=metadata['id'],
-                    doc_type=metadata['doc_type'],
                     body=jbody,
                     bulk=bulk
                 )
--- a/mq/esworker_eventtask.py
+++ b/mq/esworker_eventtask.py
@@ -143,6 +143,8 @@ def keyMapping(aDict):
         if 'utctimestamp' not in returndict:
             # default in case we don't find a reasonable timestamp
             returndict['utctimestamp'] = toUTC(datetime.now()).isoformat()
+        if 'type' not in returndict:
+            returndict['type'] = 'event'
 
     except Exception as e:
         logger.exception('Received exception while normalizing message: %r' % e)
@@ -181,7 +183,6 @@ class taskConsumer(ConsumerMixin):
             # default elastic search metadata for an event
             metadata = {
                 'index': 'events',
-                'doc_type': 'event',
                 'id': None
             }
             # just to be safe..check what we were sent.
@@ -222,14 +223,6 @@ class taskConsumer(ConsumerMixin):
             # make a json version for posting to elastic search
             jbody = json.JSONEncoder().encode(normalizedDict)
 
-            if isCEF(normalizedDict):
-                # cef records are set to the 'deviceproduct' field value.
-                metadata['doc_type'] = 'cef'
-                if 'details' in normalizedDict and 'deviceproduct' in normalizedDict['details']:
-                    # don't create strange doc types..
-                    if ' ' not in normalizedDict['details']['deviceproduct'] and '.' not in normalizedDict['details']['deviceproduct']:
-                        metadata['doc_type'] = normalizedDict['details']['deviceproduct']
-
             try:
                 bulk = False
                 if options.esbulksize != 0:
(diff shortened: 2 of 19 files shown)


mq/esworker_eventtask.py
```diff
@@ -222,14 +226,6 @@ def on_message(self, body, message):
             # make a json version for posting to elastic search
             jbody = json.JSONEncoder().encode(normalizedDict)
 
-            if isCEF(normalizedDict):
```
pwnbus: Copied comment from previous PR:

Phrozyn: This block appears unused as it is. We have no doc_types of "cef"

mq/plugins/ipFixup.py
```diff
@@ -48,12 +48,10 @@ def onMessage(self, message, metadata):
         Also sets ipv4 in two fields:
             ipaddress (decimal mapping IP)
             ipv4address (string mapping)
-            Elastic search is inconsistent about returning IPs as
```
pwnbus: Copied comment from previous PR:

Phrozyn said: facets was removed in later versions of ES after 1.0
aggregations replaced it, and this field now handles ipv4 and ipv6, though more testing is needed in ES 6 to determine if it behaves accurately. Only modified the wording here to reflect accurately.

mq/plugins/vulnerability.py
```diff
@@ -77,7 +77,7 @@ def calculate_id_v2(self, message):
         return hashlib.md5(s).hexdigest()
 
     def onMessage(self, message, metadata):
-        if metadata['doc_type'] != 'vulnerability':
+        if message['type'] != 'vulnerability':
```
pwnbus: Copied comment from previous PR:

Phrozyn said: Need to communicate this change upstream to @arcrose prior to moving to ES 6 so that the changes can be reflected in the vulnerability code that sends events to MozDef to remove doc_type from the event and add type.

mq/esworker_sns_sqs.py
```diff
@@ -162,7 +164,6 @@ def save_event(self, event, metadata):
                 self.esConnection.save_event(
                     index=metadata['index'],
                     doc_id=metadata['id'],
-                    doc_type=metadata['doc_type'],
```
pwnbus: I think we'll need to do some additional stuff here, because previously, mq plugins could and did change the doc_type  of a message, so removing this line means they no longer have that ability (which I think is fine).

 I think we need to modify the mq plugins to instead modify the `type` key on the actual message.

pwnbus: I should have added ^^ comment as part of the view, so that it did not post yet haha. I'm scrolling down and am now seeing the mq plugin changes :)

mq/esworker_cloudtrail.py
```diff
@@ -257,6 +257,10 @@ def keyMapping(aDict):
         if 'utctimestamp' not in returndict:
             # default in case we don't find a reasonable timestamp
             returndict['utctimestamp'] = toUTC(datetime.now()).isoformat()
+        if 'type' not in returndict:
```
pwnbus: I think we actually want to stomp on the type field on events here even if it exists specifically for cloudtrail events.

mq/plugins/auditdFixup.py
```diff
@@ -114,7 +114,4 @@ def onMessage(self, message, metadata):
         if 'category' not in message:
             message['category'] = 'auditd'
 
-        # set doctype
-        metadata['doc_type'] = 'auditd'
```
pwnbus: I think we still want to specifically set the type here to auditd. We don't have a reliable way to tag these events as auditd events, because while we're setting the category to 'auditd', it's an if statement, so if the incoming message already has that field set, we won't set the category as auditd.

I think we may want to either remove the about if statement to check if category exists, or we set the type to auditd and shore up the alerts that use category, to use 'type'.

mq/plugins/complianceitems.py
```diff
@@ -58,17 +58,21 @@ def onMessage(self, message, metadata):
...
             is posted with a doctype 'complianceitems'.
             Compliance items are stored in the complianceitems
-            index, with doctype last_known_state
+            index, with a type of last_known_state. The doc_type
+            is overwritten with _doc.
         """
         if not self.validate(message['details']):
             logger.error('Invalid format for complianceitem {0}'.format(message))
             return (None, None)
+        if 'type' not in message:
```
pwnbus: I don't think we want to check if `type` already exists, I think we want to automatically set that value no matter what the incoming message looks like.

mq/plugins/auditdFixup.py
pwnbus: We set message's `type` as 'auditd` here https://github.com/mozilla/MozDef/pull/1183/files#diff-a4c28c693516dbfaaf05ad3a682dfbbcR20  so whatever we decide, we'll want to sync that up as well.

mq/plugins/squidFixup.py
```diff
@@ -71,16 +73,12 @@ def onMessage(self, message, metadata):
         if message["category"] != "proxy":
             return message, metadata
 
-        # Reuse the NSM doc type
```
pwnbus: We should probably copy over these comments, since it seems weird that we're setting the type to nsm for squid events.

mq/plugins/auditdFixup.py
Phrozyn: I think if the category is auditd it should be fine, unless we want to further specify the type of auditd event, (syscall, exec, etc)

mq/esworker_cloudtrail.py
Phrozyn: Agreed.

mq/esworker_sns_sqs.py
Phrozyn: so this can be resolved yes?

mq/plugins/squidFixup.py
Phrozyn: Oh that must be leftover from when michal copied a brofixup to make this one. We should modify that type to be squid.

mq/plugins/vulnerability.py
Phrozyn: This was done and appears no upstream changes are needed.

mq/esworker_cloudtrail.py
Phrozyn: On line 222 we are converting "type" to category, which is why I put this later in the code.
How would you handle modifying this field so we aren't stomping on category?

Phrozyn: I've made a change, let me know if this is more what you were thinking.

mq/plugins/complianceitems.py
Phrozyn: I've set this to be set everytime. I had thought type was in the message already but then realized that the place where it occurs looks for it specifically several keys down, not at the top.

mq/plugins/squidFixup.py
Phrozyn: I've corrected this by setting it to squid

mq/plugins/auditdFixup.py
pwnbus: The problem is that we might not set the category to autitd if the incoming message already has a category set, because that assignment is based on a conditional at https://github.com/mozilla/MozDef/blob/ff20881548201c7aafbdde5e7b84760de1502d87/mq/plugins/auditdFixup.py#L114

So I'm saying that we might want to explicitly set type to auditd here just so we reliably have a way to "tag" incoming auditd events.

Thoughts?

Phrozyn: Is there a way for us to reliably know that "type" isn't already set to something we may find valuable that may need to be cast into another field? Or was type only added to this because maybe it was like that in the original template?

pwnbus: I'm not really sure, but my guess would be that it was added like that in the original template.

pwnbus: I just looked up and didn't see any auditd events that already have `type` set, so I think we should be fine?

Phrozyn: Sounds good!

PR #1186, opened 2019-03-26
Prep MozDef for AWS Marketplace

gene1wood:
* Enable CI/CD with AWS CodeBuild
* Create AMIs of MozDef, replicate and share them
* Publish versioned CloudFormation templates
* Link everything (container images, AMIs, templates) together by MozDef version
* RabbitMQ configured to use a real password


Phrozyn:
Not a blocking comment, but feel you might want to disable swap on the elasticsearch container. That particular service should never swap. I did create an issue for this yesterday.

https://github.com/mozilla/MozDef/issues/1189
There's a link inside that which leads to the description of how to disable swap in the container, I believe it's as simple as adding the appropriate switches to the Makefile run command.

andrewkrug:
@Phrozyn I think we'll leave it as is and consider this an optimization for the next round of work.  Thanks for the advice though.

gene1wood:
@Phrozyn And thanks for opening the GitHub issue on it so it doesn't slip through the cracks

mozdef_util/mozdef_util/elasticsearch_client.py
```diff
@@ -85,6 +85,9 @@ def create_alias_multiple_indices(self, alias_name, indices):
     def get_alias(self, alias_name):
         return self.es_connection.indices.get_alias(index='*', name=alias_name).keys()
 
+    def get_aliases(self):
```
pwnbus: Since we're modifying mozdef_util, we need to:
1. Bump the mozdef-util version here https://github.com/mozilla/MozDef/blob/master/mozdef_util/setup.py#L59 
2. Publish the new mozdef-util library to pypi (there are make targets for this in https://github.com/mozilla/MozDef/blob/master/mozdef_util/Makefile)
3. Update https://github.com/mozilla/MozDef/blob/master/requirements.txt#L34 to point to the new mozdef_util version

pwnbus: Updated in https://github.com/mozilla/MozDef/pull/1186/commits/c4f553d207839204bc46c191a1564fd9ae3ab059 and https://github.com/mozilla/MozDef/pull/1186/commits/35e9c0d4dc00b474f904dde0f133c32ccb6157bc

PR #1190, opened 2019-03-28
Replacing _type with type as this is how the kibana index presents da…

Phrozyn:
…shboards, also adding the subkey dashboard to the dashboard name parameter.

### DIFF
--- a/rest/index.py
+++ b/rest/index.py
@@ -521,12 +521,12 @@ def kibanaDashboards():
     try:
         es_client = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
         search_query = SearchQuery()
-        search_query.add_must(TermMatch('_type', 'dashboard'))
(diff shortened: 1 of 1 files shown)


rest/index.py
```diff
@@ -521,12 +521,12 @@ def kibanaDashboards():
     try:
         es_client = ElasticsearchClient((list('{0}'.format(s) for s in options.esservers)))
         search_query = SearchQuery()
-        search_query.add_must(TermMatch('_type', 'dashboard'))
+        search_query.add_must(TermMatch('type', 'dashboard'))
```
pwnbus: If you create a dashboard in the new kibana, does it set the type to `dashboard` automatically?

pwnbus: I just read the overflow of the description with the title, so I think that's a yes, that's neat!

PR #1193, opened 2019-03-29
Es6 changes eshealth

Phrozyn:
All nodes return the same stats now. We should return all node stats to the about page, and ensure we have visibility into their state too.

### DIFF
--- a/cron/healthToMongo.py
+++ b/cron/healthToMongo.py
@@ -6,43 +6,18 @@
 # Copyright (c) 2014 Mozilla Corporation
 
 
-import logging
 import requests
 import sys
 from datetime import datetime
 from configlib import getConfig, OptionParser
-from logging.handlers import SysLogHandler
 from pymongo import MongoClient
 
+from mozdef_util.utilities.logger import logger, initLogger
 from mozdef_util.utilities.toUTC import toUTC
 from mozdef_util.elasticsearch_client import ElasticsearchClient
 from mozdef_util.query_models import SearchQuery, TermMatch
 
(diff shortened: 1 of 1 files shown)


cron/healthToMongo.py
```diff
@@ -102,12 +77,6 @@ def getEsNodesStats():
     jsonobj = r.json()
     results = []
     for nodeid in jsonobj['nodes']:
-        # Skip non masters and non data nodes since it won't have full stats
```
pwnbus: Do these type of nodes now contain full stats where we no longer have to do this logic?

Phrozyn: Yes, it's been doing this since the move to 5.x

cron/healthToMongo.py
```diff
@@ -6,45 +6,20 @@
...
 from configlib import getConfig, OptionParser
-from logging.handlers import SysLogHandler
 from pymongo import MongoClient
 
+from mozdef_util.utilities.logger import logger, initLogger
 from mozdef_util.utilities.toUTC import toUTC
 from mozdef_util.elasticsearch_client import ElasticsearchClient
 from mozdef_util.query_models import SearchQuery, TermMatch
 
 logger = logging.getLogger(sys.argv[0])
```
pwnbus: We need to delete this line now that we're using mozdef_util/utilities/logger

Phrozyn: will do!

Phrozyn: done!

PR #1197, opened 2019-03-29
Update changelog for v1.38.2

pwnbus:


pwnbus:
@gene1wood sounds good, I just merged that so we should be all set.

PR #1204, opened 2019-04-03
Fixup fixup duo logpull srcip

mpurzynski:


cron/duo_logpull.py
```diff
@@ -111,10 +111,12 @@ def process_events(mozmsg, duo_events, etype, state):
 
             details[i] = e[i]
         mozmsg.set_category(etype)
-        mozmsg.details = normalize(details)
-        if "access_device" in details:
-            if "ip" in details["access_device"]:
-                mozmsg.details["sourceipaddress"] = details["access_device"]["ip"]
+        localdetails = normalize(details)
+        if "access_device" in localdetails:
+            if "ip" in localdetails["access_device"]:
```
gdestuynder: not if you're not sure of the incoming structure, you may want to do localdetails.get('access_device') and so on to avoid raising
unless you'd rather raise and fix in which case its actually better this way

PR #1207, opened 2019-04-04
This fixes the weird whiteout things that happens when moving across …

Phrozyn:
…incidents/investigations tables.


Phrozyn:
Closing in favor of #1175

PR #1214, opened 2019-04-10
Change CI/CD sequence for v1.38.5

gene1wood:
Changed the CI/CD order to now build docker images in CodeBuild, upload them
to DockerHub and then pull them down in the packer instance. Updated docs.

I confirmed this one works in a branch and built a new stack and updated an existing stack with the resulting artifacts.

### DIFF
--- a/docker/compose/mozdef_meteor/Dockerfile
+++ b/docker/compose/mozdef_meteor/Dockerfile
@@ -46,7 +46,8 @@ RUN mkdir -p /opt/mozdef/envs/meteor/mozdef
 RUN if [ "${METEOR_BUILD}" = "YES" ]; then \
         cd /opt/mozdef/envs/mozdef/meteor && \
         meteor npm install && \
-        meteor build --server localhost:3002 --directory /opt/mozdef/envs/meteor/mozdef && \
+        echo "Starting meteor build" && \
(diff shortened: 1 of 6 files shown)


docker/compose/mozdef_meteor/Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN mkdir -p /opt/mozdef/envs/meteor/mozdef
 RUN if [ "${METEOR_BUILD}" = "YES" ]; then \
         cd /opt/mozdef/envs/mozdef/meteor && \
         meteor npm install && \
-        meteor build --server localhost:3002 --directory /opt/mozdef/envs/meteor/mozdef && \
+        echo "Starting meteor build" && \
+        time meteor build --server localhost:3002 --directory /opt/mozdef/envs/meteor/mozdef && \
```
pwnbus: Do we still want to include this time command? Or was this for debugging purposes?

gene1wood: I'd hoped to include it as it's useful information during the build to show what just happened for the last 3 minutes. I looked into how to get `meteor build` to output some kind of progress our output and had no luck. So yes, I think we should keep the `time` in. Is there any downside?

pwnbus: I'm fine with keeping it in, just curious if it was intentional or not.

PR #1226, opened 2019-04-18
Add case to pass on import errors (after logging)

darakian:
My last commit in this vein just made the error apparent. This commit allows for entries to exist in the schedule which are not importable and for the alerts container to carry on without them.


Phrozyn:
This is failing due to trailing whitespace:
/alerts/celeryconfig.py:72:13: W291 trailing whitespace

darakian:
My bad. Updated.

PR #1230, opened 2019-04-23
Add logger import ipblocklist

Phrozyn:


cron/createIPBlockList.py
```diff
@@ -18,28 +18,7 @@
...
-
-
-logger = logging.getLogger(sys.argv[0])
-
-
-def loggerTimeStamp(self, record, datefmt=None):
-    return toUTC(datetime.now()).isoformat()
-
-
-def initLogger():
```
pwnbus: We need to remove https://github.com/mozilla/MozDef/blob/53d84faf252a0539a29d815bf32b416b48257119/cron/createIPBlockList.py as well

Phrozyn: Done, I think. Your link just shows the whole page :)

pwnbus: Oh wow, ha, yeah, I meant to tag the specific line :)

PR #1231, opened 2019-04-23
Paginations

jeffbryner:
An update to the UI to use the paginations component https://github.com/Kurounin/Pagination and to allow the user to set their desired page size.


Phrozyn:
Code looks good, running docker build now to QA functionality.

jeffbryner:
Interesting, if I run it via 'meteor' instead of bundling it seems to not build the indexes required for searching: 
```
Exception from sub investigations id a8o8BzNuDwLrX4APt { MongoError: text index required for $text query
errmsg: 'text index required for $text query',
```

PR #1232, opened 2019-04-24
Use wildcard in indices for searching

pwnbus:
Now that ES is more efficient with timestamped data, we can expand the search indices across all events-*, and all alerts-*. Previously, we were restricting our search window between 24 - 48 hours (events to events-previously).

We ran some benchmarks, and it turns out there isn't a very large performance hit when expanding to wildcards:
```
------------ benchmark: 2 tests -----------ons (2/2)
Name (time in ms)            Mean          
-------------------------------------------
test_events_weekly        90.7345 (1.0)    
test_events_wildcard     105.5318 (1.16)   
-------------------------------------------
```

### DIFF
--- a/alerts/lib/alerttask.py
+++ b/alerts/lib/alerttask.py
@@ -106,7 +106,10 @@ class AlertTask(Task):
         self._configureKombu()
         self._configureES()
 
-        self.event_indices = ['events', 'events-previous']
+        # We want to select all event indices
+        # and filter out the window based on timestamp
+        # from the search query
+        self.event_indices = ['events-*']
 
     def classname(self):
         return self.__class__.__name__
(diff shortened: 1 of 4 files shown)


pwnbus:
We'll have to release a new version of mozdef_util as we changed the default indices.

jeffbryner:
Any impact to memory utilization?

Phrozyn:
We should take a look at that. JVM usage could be increased by this method. Good call, I didn't think of that.

pwnbus:
I took a look at CPU, memory (via top) and the JVM during the testing and it didn't seem to have any noticeable difference.

jeffbryner:
> I took a look at CPU, memory (via top) and the JVM during the testing and it didn't seem to have any noticeable difference.

OK, it's likely they've gotten much better at only sending queries around to places that are likely to have hits + JVM would increase for hits rather than misses.

PR #1235, opened 2019-04-25
Update docs to add alert actions and plugins

pwnbus:


docs/source/actions.rst
```diff
@@ -0,0 +1,38 @@
+Actions
+=======
+
+Actions are supported in the end of the alert pipeline currently.
```
Phrozyn: change "in" to "at"

pwnbus: I like that better, added in https://github.com/mozilla/MozDef/pull/1235/commits/9d45ad5415db037ad647d01a0f4049f6bb0a6330

Issue #1246, opened 2019-05-03
Batching log input


darakian:
Hey all,

Is it possible to batch log input into MozDef? Looking at the docs here
https://mozdef.readthedocs.io/en/latest/usage.html
I don't see any mention of it. I've tried bundling my logs as my desired format of a json array of messages
`[{msg1}, {msg2}, ...]` and the input gets rejected.
In the code I see 
```
@route('/_bulk',method='POST')
@route('/_bulk/',method='POST')
def bulkindex():
```
and sending logs to `mozdef:8080/_bulk` gives back a 200, but I don't see my logs appearing. Is this an under construction section of the code or am I misreading things?

Edit:
Data format sent to `_bulk`
`'[{"summary": "just a double test"}, {"summary": "just a double test"}]'`

Phrozyn:
The loginput code may require each event on a single line rather than separated by a comma. 

The first thing the script does is add each event line by line to an "eventlist" list.

Otherwise it may hit a character limit by attempting to index every event as a single event. 

You should have a log output called loginput.log wherever you've directed mozdef to write logs. 

Do you see anything in the log?

darakian:
Ah I see. I'll test with newlines and check my logs on monday. For my log output it may be easier to send comma-space separated values. Would you be opposed to enabling general white space separated arrays or perhaps a configurable option? I don't mind making the PR for this if so.

darakian:
PR here
https://github.com/mozilla/MozDef/pull/1247

Phrozyn:
Merged #1247
Closing

PR #1247, opened 2019-05-06
Add json array input decode with try catch to fall back to string splitting

darakian:
This PR adds a bulk message decode route compatible with rsyslog's jsonarray output. In my testing all that was required was to decode the raw message which was nice. Discussion on the subject started here
https://github.com/mozilla/MozDef/issues/1246

I'm not sure what the `# don't post the items telling us where to post things..` part was about so I just copied over the publishing logic.

Edit: Rsyslog docs for http batch formats
https://www.rsyslog.com/doc/v8-stable/configuration/modules/omhttp.html#batch-format

### DIFF
--- a/loginput/index.py
+++ b/loginput/index.py
@@ -41,9 +41,21 @@ def bulkindex():
         bulkpost=request.body.read()
         # bottlelog('request:{0}\n'.format(bulkpost))
         request.body.close()
-        if len(bulkpost)>10:  # TODO Check for bulk format.
-            # iterate on messages and post to event message queue
+        try:  # Handles json array bulk format [{},{},...]
+            messages = json.loads(bulkpost)
+            for event in messages:
+                # don't post the items telling us where to post things..
+                if 'index' not in event:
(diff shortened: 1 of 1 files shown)


darakian:
Can I get a comment on if this behavior is desired?

pwnbus:
@darakian I think the behavior is desired, but looking quickly, won't this post the bulk messages twice?

darakian:
Oh shit you're right. I guess it should return after the first post. No worries on the delay, I just wanted to be sure that I'm not working against the grain. Give me a bit and I'll update the PR.

darakian:
@pwnbus Just pushed an update. I moved a few comments around in hopes that it makes things a bit clearer for the next person. Let me know if you agree.

Phrozyn:
@darakian lines 44 and 56 are failing formatting here:

/loginput/index.py:44:13: E261 at least two spaces before inline comment
./loginput/index.py:56:29: E261 at least two spaces before inline comment

If you can fix those, this should pass.
Thanks!

darakian:
@Phrozyn Pushed an update. My bad on the whitespace.

PR #1248, opened 2019-05-06
Fixup mozdefutil es6

pwnbus:


mozdef_util/mozdef_util/elasticsearch_client.py
```diff
@@ -160,6 +166,9 @@ def save_event(self, body, index='events', doc_id=None, bulk=False):
         event.add_required_fields()
         return self.__save_document(index=index, body=event, doc_id=doc_id, bulk=bulk)
 
+    def save_dashboard(self, body, dash_name, doc_id=None):
+        return self.es_connection.index(index='.kibana', doc_type='dashboard', body=body, id=doc_id)
```
Phrozyn: this is incorrect for es6, it should be 'doc' the "type" is set to 'dashboard'
Otherwise, when using kibana and attempting to save a new dashboard through the interface, this is going to cause probs.

tests/mozdef_util/test_elasticsearch_client.py
```diff
@@ -28,11 +28,11 @@ def setup(self):
     def get_num_events(self):
         self.refresh('events')
         search_query = SearchQuery()
-        search_query.add_must(TermMatch('_type', '_doc'))
+        search_query.add_must(TermMatch('_type', DOCUMENT_TYPE))
```
Phrozyn: Is this needed since it will always be 1 for a particular index?
Should this be for "type" instead or some other field?

pwnbus: This is basically just a utility helper to return the number of events in an index, so that tests can verify that they were able to write events to ES.

mozdef_util/mozdef_util/elasticsearch_client.py
pwnbus: I ended up removing this function all together as one rest api unt test was the only thing using it.

tests/mozdef_util/test_elasticsearch_client.py
Phrozyn: ahh makes sense then :)

tests/rest/ssh_dashboard.json
```diff
@@ -1,285 +1,16 @@
...
-    "filter": {
-      "list": {},
-      "ids": []
+  "dashboard": {
+    "title": "Your Basic Dashboard",
+    "hits": 0,
+    "description": "",
+    "panelsJSON": "[]",
+    "optionsJSON": "{\"darkTheme\":false,\"useMargins\":true,\"hidePanelTitles\":false}",
+    "version": 1,
```
Phrozyn: version should be removed, this has caused problems in the past as kibana likes to control version itself, it will automatically assign the version on it's own.

tests/rest/ssh_dashboard.json
```diff
@@ -1,285 +1,16 @@
...
-      "ids": [
-        0
-      ]
-    },
-    "filter": {
-      "list": {},
-      "ids": []
+  "dashboard": {
+    "title": "Your Basic Dashboard",
+    "hits": 0,
```
Phrozyn: hits should be removed as this also causes problems, kibana keeps track of hits on it's own and will autopopulate this field.

tests/rest/ssh_dashboard.json
pwnbus: Good catch, I'll remove it!

tests/rest/ssh_dashboard.json
pwnbus: Good catch, removed!

PR #1251, opened 2019-05-07
Multiple Cloud Fixes

andrewkrug:
* Support NoSSL as an option
* Support Basicauth as an option
* Support OIDC using the authenticating ALB
* Configure log rotate for all docker containers to prevent disk from filling
* Introduce MozDef Cognito Proxy to handle ALB headers and support cognito configuration
* Fix ElasticSearch Sec group rules preventing customer deploys.
* Improve helper text via the onboarding wizard.


andrewkrug:
Closed in favor of an inbound pull that adds SQS support and addresses the requested change.

PR #1253, opened 2019-05-08
IP address enrichment

arcrose:
This Pull Request is a work in progress.  I am opening it early to start getting some feedback and guidance.

The goal of this PR is to add a new alert plugin that will apply to all alerts.  Upon finding IPv4 or IPv6 addresses in any part of an alert's values, it will reference a configured set of "known IPs" that provide text to enrich the alert with.  This "enriched" text will be appended to the summary of the alert, so that it can be read by people without potentially breaking any other components by changing alert fields.


Phrozyn:
per travis:
./alerts/plugins/ip_source_enrichment.py:26:7: E121 continuation line under-indented for hanging indent

arcrose:
Question for everyone:

Rather than appending information to the summary field, should we create something like a `details.office` field?

alerts/plugins/ip_source_enrichment.py
```diff
@@ -0,0 +1,114 @@
...
+        ip_address = netaddr.IPAddress(ip)
+
+        matching_descriptions = filter(
+            lambda known: ip_address in netaddr.IPSet([known['range']]),
+            known_ips)
+
+        for desc in matching_descriptions:
+            enriched = desc['format'].format(ip)
+
+            alert['summary'] += '; ' + enriched
```
tristanweir: Instead of just adding to the summary, we should set a data field in the alert as well. This will allow for easier filtering, keying, etc.

Ex: `details.iplocation: OFFICE2`

alerts/plugins/ip_source_enrichment.json.conf
```diff
@@ -0,0 +1,14 @@
+{
+  "known": [
+    {
+      "ipVersion": 4,
+      "range": "8.32.0.0/16",
+      "format": "{0} is in OFFICE1."
```
tristanweir: Instead of having a single human-readable string containing the data item, it may be better to have a data item for `location`. Easier to parse if needed.

So:
```
      "ipVersion": 4,
      "range": "8.32.0.0/16",
      "location:" "OFFICE1",
      "format": "{1} is in {2}."
```

alerts/plugins/ip_source_enrichment.py
```diff
@@ -0,0 +1,114 @@
...
+from operator import add
+import os
+import re
+
+import netaddr
+
+
+CONFIG_FILE = os.path.join(
+    os.path.dirname(__file__),
+    'ip_source_enrichment.json.conf')
```
pwnbus: The convention we have is to name the file simply `ip_source_enrichment.json` if it's a json formatted config file.

alerts/plugins/ip_source_enrichment.py
```diff
@@ -0,0 +1,115 @@
...
+    The expected format of the configuration file,
+    `ip_source_enrichment.json.conf`, is as follows:
+
+    ```json
+    {
+      "known": [
+        {
+          "range": "1.2.3.4/8",
+          "site": "office1",
+          "format": "IPv4 {1} is known"
```
tristanweir: This should ref `{0}`. Same with the entry below. I know it's just a comment but we want it to be consistent

alerts/plugins/ip_source_enrichment.py
```diff
@@ -0,0 +1,115 @@
...
+    alert = alert.copy()
+
+    for ip in set(ips):
+        ip_address = netaddr.IPAddress(ip)
+
+        matching_descriptions = filter(
+            lambda known: ip_address in netaddr.IPSet([known['range']]),
+            known_ips)
+
+        for desc in matching_descriptions:
```
tristanweir: What happens when both the sourceipaddress and destinationipaddress are in the known ranges? I think we would overwrite details.site with whatever came last and append both to the summary. Is this what 'should' happen?

arcrose: This is a good question and part of why I didn't add anything like `details.site` the first time around. I defer to @Phrozyn or @pwnbus or @mpurzynski regarding how we should handle the case where multiple IPs are recognized.

We could do something like:

```json
{
  "details": {
    "sites": [
      { "ip": "a.b.c.d", "site": "office1" },
      { "ip": "x.y.z.n", "site": "office2" }
    ]
  }
}
```

tristanweir: This approach works for me

alerts/plugins/ip_source_enrichment.json
```diff
@@ -0,0 +1,14 @@
+{
+  "known": [
+    {
+      "range": "8.32.0.0/16",
+      "site": "OFFICE1",
+      "format": "{0} is in OFFICE1."
```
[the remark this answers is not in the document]
Phrozyn: something like "format": "{0} is in {1}." where {1} is "site" value?

Phrozyn:
> Question for everyone:
> 
> Rather than appending information to the summary field, should we create something like a `details.office` field?

https://github.com/mozilla/MozDef/blob/master/mq/plugins/mozilla_location.py#L34
This is how we currently desgnate office locations. Did you work from this?

alerts/plugins/ip_source_enrichment.json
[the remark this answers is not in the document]
pwnbus: I don't see the benefit of including a formatted string, as long as the data we care about is stored in "site"?

alerts/plugins/ip_source_enrichment.py
pwnbus: I'd agree, having details.sites be an array containing all of the different ips -> office mappings.

alerts/plugins/ip_source_enrichment.json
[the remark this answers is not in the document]
arcrose: I think there are bits of info we'd like to include that might not make sense to put in 'site' or somewhere else.  It might just be some really specific description text

alerts/plugins/ip_source_enrichment.py
```diff
@@ -0,0 +1,146 @@
...
+
+            alert['details']['sites'].append({
+                'ip': ip,
+                'site': desc['site'],
+            })
+
+    return alert
+
+
+def _load_config(file_path):
```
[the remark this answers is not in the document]
arcrose: Well, the _point_ of the function is to load the alert plugin's configuration. I hoped that would be obvious by:

1. The name
2. Its usage context and
3. Its doc string

To answer the question you didn't ask:

I left function in its sort of half-baked state because I wasn't sure what the recommended way to open up a config file is.  Normally this would be a basic thing but I know the alert plugins are run on workers and might have a different base path from another component.

Would an implementation as simple as

```py
try:
  f = open(file_path)
  return json.load(f)
except IOError:
  return { 'known': [] }
```

be okay?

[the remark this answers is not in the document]
pwnbus: Yeah, that looks good from a config loading perspective.

PR #1256, opened 2019-05-09
Fixup watchlist alert to have auth be configurable

pwnbus:


alerts/get_watchlist.conf
```diff
@@ -2,3 +2,4 @@
 # set the following to your protected endpoint api_url
 api_url = http://localhost:8081/getwatchlist
 jwt_secret = secret
+use_auth = False
```
tristanweir: In get_watchlist.py, we do
`         if self.config.use_auth.lower() != 'false': `

Even though we lowercase the flag anyway, do we want to set the initial flag as lowercase since that's what we expect in code?

pwnbus: Yeah, makes sense. Added in the newest commit.

PR #1259, opened 2019-05-10
Alert sqs support

andrewkrug:
Just PRing to test docker support for --parallel in travis


alerts/lib/alerttask.py
```diff
@@ -109,56 +109,86 @@ def __init__(self):
...
+            connString = self.__build_conn_string()
             self.mqConn = kombu.Connection(connString)
-
-            self.alertExchange = kombu.Exchange(
-                name=RABBITMQ['alertexchange'],
-                type='topic',
-                durable=True)
-            self.alertExchange(self.mqConn).declare()
-            alertQueue = kombu.Queue(RABBITMQ['alertqueue'], exchange=self.alertExchange)
+            if connString.find('sqs') == 0:
```
pwnbus: Since you're creating the string on `__build_conn_string`, would it be better to verify if the string starts with "sqs://" instead of if `sqs` is in the connection string?

PR #1262, opened 2019-05-13
Fixup kibanaurl for cloudy mozdef setup

pwnbus:


cloudy_mozdef/cloudformation/mozdef-instance.yml
```diff
@@ -122,9 +122,7 @@ Resources:
                 # Drives the configuration of variables for a variety of containers.
                 OPTIONS_ESSERVERS=${ESURL}
                 OPTIONS_KIBANAURL=${KibanaURL}
-                # The OPTIONS_METEOR_KIBANAURL uses the reserved word "relative" which triggers MozDef
-                # to use relative links to Kibana : https://github.com/mozilla/MozDef/pull/956
-                OPTIONS_METEOR_KIBANAURL=https://relative:9090/_plugin/kibana/
+                OPTIONS_METEOR_KIBANAURL=$OPTIONS_KIBANAURL
```
pwnbus: Although I wonder if this needs to be `OPTIONS_METEOR_KIBANAURL=${OPTIONS_KIBANAURL}`

andrewkrug: This is cloudformation syntax... for variable expansion.  Not bash expansion

gene1wood: @pwnbus Ya, you'll want to add some curly braces. [More information on this CloudFormation Fn::Sub syntax](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html)

gene1wood: But I think this has been fixed as I can't find this in the current codebase of this PR

PR #1264, opened 2019-05-14
Create simple cloudtrail excessive describe calls alert

pwnbus:


tests/alerts/alert_test_suite.py
```diff
@@ -173,7 +173,6 @@ def verify_saved_events(self, found_alert, test_case):
         if self.deadman:
             return
 
-        assert len(found_alert['_source']['events']) == len(test_case.full_events)
```
pwnbus: I'm removing this since this check doesn't work if the sample event number != threshold, now that I think about it, it might not be providing much value in the first place.

alerts/cloudtrail_excessive_describe.py
```diff
@@ -0,0 +1,37 @@
...
+class AlertCloudtrailExcessiveDescribe(AlertTask):
+    def main(self):
+        # Create a query to look back the last 20 minutes
+        search_query = SearchQuery(minutes=20)
+
+        # Add search terms to our query
+        search_query.add_must([
+            TermMatch('source', 'cloudtrail'),
+            TermMatch('details.eventverb', 'Describe'),
+            ExistsMatch('details.source')
```
gene1wood: It looks like using `details.source` as the aggregation field doesn't work. This field is only present in autoscaling events which probably isn't what was intended.

I'm not entirely sure what was envisioned for this alert, if it was to detect when a single threat actor was making many describe calls, or when a single AWS service (e.g. DynamoDB) was receiving many describe calls (regardless of who's making them), or when an AWS account as a whole is receiving many describe calls (again, unrelated to who's making them)

If I can figure out what the intent is I can suggest what field should be used for aggregation.

As it is right now this only detects when AWS autoscaling service makes too many describe calls.

PR #1270, opened 2019-05-19
Reinforce2019

andrewkrug:


alerts/lib/config.py
```diff
@@ -31,9 +32,14 @@
     'alertqueue': 'mozdef.alert'
 }
 
-ES = {
-    'servers': ['http://localhost:9200']
-}
+if os.getenv('OPTIONS_ESSERVERS'):
```
pwnbus: I think this logic can be a bit more legible if it was something like:
```
es_server = "http://localhost:9200"
if os.getenv('OPTIONS_ESSERVERS'):
    es_server = os.getenv('OPTIONS_ESSERVERS')
ES = {
    'servers': [es_server]
}
```

PR #1277, opened 2019-05-21
Basic auth password should be randomized

gdestuynder:
https://jira.mozilla.com/browse/EIS-970

> Currently in order to prototype I set the basic auth password for a user to be mozde:mozdef.  Let's randomize this and output it as a parameter that the user can get to in cloudformation console instead.

Currently testing this


gdestuynder:
Note: I did not do the more extensive testing that I planned for this PR. I've marked it as read for review so that Andrew could see it and merge it if necessary

gdestuynder:
@gene1wood we decided against randomization but instead have the user set it..

PR #1278, opened 2019-05-21
[WIP] Initial commit of a sample alert work in progress for cloudtrail login.

Phrozyn:
This doesn't pass unit tests yet, which complain that the alert classname doesn't match the expected name. Once this is working as expected for something simple like detecting a login, I can expand it to include other criteria to make it more actionable.


Phrozyn:
Closing this as it will not get done before the end of this half.

Issue #1027, opened 2018-12-30
Present CIDR blocks uniquely in ipblocklist.txt


gcoxmoz:
https://github.com/mozilla/MozDef/blob/master/cron/createIPBlockList.py appears to be the script that generates ipblocklist.txt, which is ingested to provide blocklist services.

We have come across a situation where you are providing (example) 1.2.3.4/24 and 1.2.3.5/24.  Those are effectively equal.

On the ingestion side, ipset(8) fails out when presented with a duplicate, which these are.  While we should put in a filter on the receiving side to prevent this from happening, it is worth considering how you want to manage CIDR blocks that you send, in order to manage these blocks' presentation in the blocklist.

ryancee:
Feedback from @pwnbus, thinking something along the lines of, https://gist.github.com/ryancee/effa5206032edacffadc266d1f46110b

pwnbus:
I believe having individual ips with a /24 is usually a mistake, so we don't really care about the host bit.

PR #1282, opened 2019-05-22
Call cidr_merge on ipblocklist ips

pwnbus:


pwnbus:
https://github.com/mozilla/MozDef/issues/1027

PR #1285, opened 2019-05-24
CICD in CloudFormation

gene1wood:
Initial capture of the CodeBuild configuration and role in CloudFormation

This likely still requires the one time binding action to link the CodeBuild project with the GitHub project and webhook. But this may get us a large part of the way there before that manual step.


pwnbus:
@gene1wood is this PR still needed?

gene1wood:
Yes. Now that reinforce is over I'll set this up for review.

Issue #1286, opened 2019-05-25
Watchlist errors out in Docker Build


Phrozyn:
[2019-05-25 00:50:00,100: ERROR/ForkPoolWorker-2] get_watchlist.AlertWatchList[26740e2d-8257-4a4a-91a6-2453ca4a5ed8]: Exception in main() method: process_alert() takes exactly 2 arguments (1 given)
Traceback (most recent call last):
  File "/opt/mozdef/envs/mozdef/alerts/lib/alerttask.py", line 535, in run
    self.main(*args, **kwargs)
  File "/opt/mozdef/envs/mozdef/alerts/get_watchlist.py", line 34, in main
    self.process_alert()
TypeError: process_alert() takes exactly 2 arguments (1 given)

Phrozyn:
Also the modal window is missing text, for the Entry boxes as well as the plugin selector.

Phrozyn:
This appears to be the case on a full install as well, not just docker.

Phrozyn:
The summary of the watchlist alert also is just "watchlist"
Will try doing a bit more troubleshooting.

PR #1287, opened 2019-05-25
Fix watchlist process_alert function

pwnbus:


pwnbus:
https://github.com/mozilla/MozDef/issues/1286

PR #1292, opened 2019-05-28
Ensure that geolite DB is fetched when MozDef launches

gene1wood:
Adds conditional fetching behavior to only fetch if the geolite DB is absent from disk
Adds crontab entry to fetch if absent when crond launches
Relates to https://jira.mozilla.com/browse/EIS-1005

### DIFF
--- a/cron/update_geolite_db.py
+++ b/cron/update_geolite_db.py
@@ -49,6 +49,8 @@ def save_db_data(save_path, db_data):
 def main():
     logger.debug('Starting')
     logger.debug(options)
+    if options.only_update_if_absent and os.path.isfile(options.db_location):
+        return True
     db_data = fetch_db_data(options.db_download_location)
     save_db_data(options.db_location, db_data)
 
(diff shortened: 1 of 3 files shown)


cron/update_geolite_db.py
```diff
@@ -49,6 +49,8 @@ def save_db_data(save_path, db_data):
 def main():
     logger.debug('Starting')
     logger.debug(options)
+    if options.only_update_if_absent and os.path.isfile(options.db_location):
```
pwnbus: I'm not sure we want to actually do this, I think we want to periodically update this file daily even if the file exists, because the ip -> country mapping can change and get out of date.

gene1wood: This conditional triggers when *both* `options.only_update_if_absent` is `True` *and* the file exists. `options.only_update_if_absent` is a new command line argument which is used in the `@reboot` line of the `cron-entries.txt`.

As a result, this doesn't relate to the normal operation of the cron job, only the first run when the system comes up.

gene1wood:
@pwnbus is going to add this behavior to the [`mozdef_bootstrap`](https://github.com/mozilla/MozDef/blob/master/docker/compose/mozdef_bootstrap/Dockerfile) container so this isn't needed.

PR #1294, opened 2019-05-28
Port scan enrichment

arcrose:


alerts/plugins/port_scan_enrichment.json
```diff
@@ -0,0 +1,13 @@
+{
+  "elasticSearchAddress": "http://127.0.0.1:9200",
+  "indicesToSearch": [
+    "events-*"
```
Phrozyn: If your search window is the last 24 hours, you should use the events-weekly alias, rather than events-*

arcrose: This is resolved in bd3d2b

alerts/plugins/port_scan_enrichment.py
```diff
@@ -0,0 +1,206 @@
...
+    ])
+
+    results = search_fn(search_query)
+
+    events = [
+        hit.get('_source', {})
+        for hit in results.get('hits', [])
+    ]
+
+    alert = alert.copy()
```
pwnbus: I don't think we need to do the copy here, the expectation is that plugins will modify the alert object directly

tests/alerts/plugins/test_port_scan_enrichment.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_alert_enriched(self):
+        results = {
+            'hits': [
+                {
+                    '_source': {
+                        'details': {
+                            'destinationipaddress': '1.2.3.4',
+                            'destinationport': 80
+                        },
+                        'timestamp': '30 minutes ago'
```
pwnbus: Since the timestamp will be ISO 8601 formatted, can we have the sample data be the same? IE: `2016-07-13 22:33:31.625443+00:00`

alerts/plugins/port_scan_enrichment.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Returns a modified alert based on a copy of the original.
+    '''
+
+    search_query = SearchQuery(**search_window)
+
+    search_query.add_must([
+        TermMatch('category', 'bro'),
+        TermMatch('source', 'conn'),
+        PhraseMatch(
```
pwnbus: I think we want to use a TermMatch here, since details.sourceipaddress will be only the actual IP with no other text, we can do an exact match essentially.

PR #1297, opened 2019-05-29
[WIP] Add wait-for-port and wait-for-container scripts in docker env

pwnbus:


gene1wood:
So there's two cases here
1. Container A needs to wait for Container B to start, run and stop before starting
2. Container A needs to wait for Container B to start, run and reach a ready state before starting

My suggestion about using `healthcheck` addresses Case 2. My example isn't useful because I used the `mozdef_bootstrap` container which is case 2.

So what about this
* We establish healthchecks for all Case 2 containers. This is probably good to do just in general so that the containers emit to docker their health
* We, for now, use the `depends_on` condition for Case 2 containers. This goes away with docker compose v3 but the healthchecks will still be valuable in v3 though we'll need to change from using `depends_on`
* For Case 1 containers, we make Container B emit the fact that it's completed by one of these two methods
  * healthcheck method
    * Add a healthcheck to Container B that looks for the presence of a `/done` file
    * At the end of the CMD, we `touch /done && sleep 120` which creates the done file and then waits for 2 minutes (longer than any other depends_on condition intervals)
    * Add depends_on conditions to Container A which depend on Container B
  * volume method
    * Container A and B we add a volume mount to a directory on the host system (the same directory)
    * On Container B at the end of the CMD, we `touch /path/to/volume/mount/done` which creates the done file
    * On Container A we do the same kind of wait loop in this PR but instead of waiting for ping to fail (which sounds sketchy) we wait for the file to appear in the filesystem

pwnbus:
Talked it over with Gene and we decided to close this PR, and will go another direction if/when we need to solve this problem. I've also created https://github.com/mozilla/MozDef/issues/1312 to add healthchecks to docker environment.

PR #1311, opened 2019-06-04
Replace "unset" with empty string

gene1wood:
Clarify that "unset" isn't a reserved word
Use an empty string as a default to make it easier to understand
Add Rules which validate
* that the user is either choosing basic auth or OIDC but not both.
* that all OIDC parameters are filled out if choosing OIDC.
* that ACM cert is filled out if choosing OIDC


gene1wood:
I've tested the various conditions that these rules test for to confirm that they allow valid parameter combinations and disallow invalid ones

gene1wood:
I'll close this and resubmit from a branch under mozilla/MozDef instead of my fork

PR #1317, opened 2019-06-06
Update urllib dependency

pwnbus:


cron/auth02mozdef.py
```diff
@@ -11,19 +11,11 @@
 import sys
 import os
 import requests
-import mozdef_client as mozdef
-from mozdef_util.utilities.dot_dict import DotDict
-
-try:
-    import urllib.parse
```
pwnbus: We weren't actually using urllib here at all, as we use requests instead.

Issue #1312, opened 2019-06-04
Add healthchecks to docker environment


pwnbus:
We should add healthchecks to each docker service, so orchestration tools know when a service is "healthy".

PR #1318, opened 2019-06-06
Add healthchecks for docker containers

pwnbus:
https://github.com/mozilla/MozDef/issues/1312


pwnbus:
@gene1wood I've created healthchecks for all containers except alerts, alertactions, cron, and mq_worker. All of those containers are main processes that don't expose any network port, so I'm wondering what the best way to check for those are, simply a `ps aux | grep `?

gene1wood:
> All of those containers are main processes that don't expose any network port, so I'm wondering what the best way to check for those are, simply a ps aux | grep ?

@pwnbus I'd say the best thing to do is to extend those containers to assert their health in the filesystem. For example, take cron. You could make each cron job emit it's exit status on each run into a file in the filesystem (or files), then have your healthcheck report based on the aggregate values from all cron jobs. That way the container says it's healthy if all of it's constituent components are health.

PR #1325, opened 2019-06-12
Removes the update_all_types call from mozdef-util

Phrozyn:
This functionality is no longer needed in 6.x as there is only one type.
Leaving this in the abstraction layer will cause the rotation of indices to fail.


Phrozyn:
Closing this as it's already in the PR, just need to update qa2's instance of mozdef-util

PR #1330, opened 2019-06-14
Removing deprecated "region" parameter

Phrozyn:


Phrozyn:
this is also apparently merging master into the breaking es6 changes branch

PR #1333, opened 2019-06-21
Fix cloudtrail_excessive_describe alert

gene1wood:
Change aggregation field since details.source only exists for autoscaling events
This should fix the bug in #1264


Phrozyn:
@gene1wood if you can resolve the conflicts we can get this reviewed! thank you!

gene1wood:
It looks like @andrewkrug fixed this in [these commits](https://github.com/mozilla/MozDef/commits/reinforce2019/alerts/cloudtrail_excessive_describe.py) so I'll close this PR unmerged

PR #1340, opened 2019-06-28
Ensure dot notation for local package imports

pwnbus:


pwnbus:
This will require a new version of mozdef_util to be created, with support still for python2.

PR #1342, opened 2019-06-28
Python 3 upgrade

pwnbus:


requirements.txt
```diff
@@ -25,15 +23,15 @@ google-api-python-client==1.4.0
...
+ipaddr==2.2.0
 ipaddress==1.0.17
 ipwhois==0.15.0
 jmespath==0.9.3
 kombu==4.1.0
 meld3==1.0.2
 mozdef-client==1.0.11
-mozdef-util==2.0.3
-netaddr==0.7.1
+mozdef-util==3.0.0.dev0
```
pwnbus: I need to make sure this gets changed, IE: do a mozdef-util release.

mozdef_util/HISTORY.rst
```diff
@@ -79,3 +79,10 @@ Add is_ip utility function
 ------------------
 
 * Fixed static file includes in python package
+
+
+3.0.0 (2019-??-??)
```
pwnbus: And will need to update this too.

requirements.txt
pwnbus: Done, created 3.0.0

PR #1358, opened 2019-07-08
Fix bulk queue to lock before saving to ES

pwnbus:


pwnbus:
FYI the unit tests will fail, and that's expected.

PR #1371, opened 2019-07-16
Improve Dockerfiles

gene1wood:
* Change pattern of a final recursive chown to instead intentionally setting
  the owner and group in the COPY commands. This should avoid a layer that
  touches all previously copied files
* Add checks for yum repo signing key fingerprint
* Expand some command line arguments to long form so it's easily
  understandably by the reader without having to lookup the man page
* Add removal of yum cache to reduce docker image size
* Change kibana WORKDIR to a fixed and simple value of /kibana
* Cleanup long lines to be more readable
* Use symbolic links instead of internal file copies to reduce docker image size
* Standardize on indents of two spaces

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ run-tests-resources:  ## Just run the external resources required for tests
 .PHONY: run-test
 run-test: run-tests
 
-.PHONY: run-test
+.PHONY: run-tests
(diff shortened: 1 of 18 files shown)


gene1wood:
Heh, conflicts just within the last few hours. I'll fix and resubmit a PR

PR #1377, opened 2019-07-17
Update healthToMongo to include 2 digits for disk usage

pwnbus:
This fixes the issue of disk usage and total having a bunch of decimal places, we only now include 2.


pwnbus:
This also rounds the EPS values in the about page as well.

PR #1381, opened 2019-07-18
Implement separate function for getting open indices

pwnbus:
This also creates mozdef_util version 3.0.3.


pwnbus:
The tests will fail until we publish a new version of mozdef_util.

Issue #1380, opened 2019-07-18
Move to boto3 breaks sqs and CloudTrail workers due to empty region value


gene1wood:
With the addition of #1342 the mq worker used in our cloud deployment for both CloudTrail and SQS no longer works.

This is because we're passing a blank region value instead of inheriting it from the environment. The result are errors like

```
2019-07-17T23:52:02.232762+00:00 - esworker_cloudtrail.py - INFO - started without uwsgi
Traceback (most recent call last):
  File "esworker_cloudtrail.py", line 452, in <module>
    main()
  File "esworker_cloudtrail.py", line 390, in main
    options.secretkey)
  File "../mq/lib/sqs.py", line 15, in connect_sqs
    **credentials
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/boto3/__init__.py", line 100, in resource
    return _get_default_session().resource(*args, **kwargs)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/boto3/session.py", line 389, in resource
    aws_session_token=aws_session_token, config=config)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/session.py", line 885, in create_client
    client_config=config, api_version=api_version)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/client.py", line 76, in create_client
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/client.py", line 285, in _get_client_args
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/args.py", line 45, in get_client_args
    endpoint_url, is_secure, scoped_config)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/args.py", line 111, in compute_client_args
    service_name, region_name, endpoint_url, is_secure)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/client.py", line 358, in resolve
    service_name, region_name)
  File "/opt/mozdef/envs/python/lib/python3.6/site-packages/botocore/regions.py", line 122, in construct_endpoint
    partition, service_name, region_name)
[...]

PR #1383, opened 2019-07-18
Fix region parameter in connect_sqs

pwnbus:
This "might" be a fix for https://github.com/mozilla/MozDef/issues/1380


pwnbus:
FYI this is untested, so I'd like to see if we could test this before merging perhaps?

I also tried to see if we needed to modify the calls to access S3 buckets via regionless parameters, but we may not need to make any modifications? (not entirely sure)

pwnbus:
We may need to do a similar patch to https://github.com/mozilla/MozDef/blob/master/mq/esworker_cloudtrail.py#L231-L235

pwnbus:
Verified that this is working, so it's ready to review/merge.

pwnbus:
@gene1wood it wasn't a breaking change since we were passing parameters via position when calling that, but I've updated it to be consistent in https://github.com/mozilla/MozDef/pull/1383/commits/d7437b46197a7cf92f8892dbb4d5a59515ef4d6e

gene1wood:
Aha right, [it's called positionally](https://github.com/mozilla/MozDef/pull/1383/files#diff-252fe61df1b95fcce77d038d7838f7f2R9) got it. Cool, LGTM

PR #1387, opened 2019-07-19
Update Elasticsearch version for cloudymozdef to 6.8

pwnbus:


pwnbus:
@andrewkrug FYI

pwnbus:
We had to downgrade this to 6.7 since 6.8 wasn't an available possible Managed ES version.

PR #1388, opened 2019-07-21
fix for plugins, needs more work.

Phrozyn:
ipblocklist is working with this, as well as watchlist, but ipblocklist is still throwing an indexing error and I can't seem to see the dict that it's trying to index.

### DIFF
--- a/rest/plugins/fqdnblocklist.py
+++ b/rest/plugins/fqdnblocklist.py
@@ -205,18 +205,18 @@ class message(object):
         try:
             for i in request.json:
                 # were we checked?
-                if self.name in i:
-                    blockfqdn = i.values()[0]
-                if 'fqdn' in i:
-                    fqdn = i.values()[0]
-                if 'duration' in i:
-                    duration = i.values()[0]
-                if 'comment' in i:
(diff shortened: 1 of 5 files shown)


rest/plugins/fqdnblocklist.py
```diff
@@ -205,18 +205,18 @@ def onMessage(self, request, response):
...
-                    fqdn = i.values()[0]
-                if 'duration' in i:
-                    duration = i.values()[0]
-                if 'comment' in i:
-                    comment = i.values()[0]
-                if 'referenceid' in i:
-                    referenceID = i.values()[0]
-                if 'userid' in i:
-                    userid = i.values()[0]
+                if self.name in list(i.keys()):
```
pwnbus: These changes aren’t needed, you can look for if a key exists in python3 the same as python2

Phrozyn: removed .keys() from the fqdnblocklist.py

pwnbus:
Closing in favor of https://github.com/mozilla/MozDef/pull/1394

PR #1389, opened 2019-07-22
Updating some directory structure for manual installs.

Phrozyn:


docs/source/installation.rst
```diff
@@ -110,7 +110,7 @@ Then::
 
   source /opt/mozdef/envs/python/bin/activate
   cd /opt/mozdef/envs/mozdef
-  PYCURL_SSL_LIBRARY=nss pip install -r requirements.txt
```
pwnbus: By this time, we’re already in the virtual environment, so pip will be “pip”, not pip3.

Phrozyn: I'll need to re-test this because I ran into errors. I had to uninstall pycurl, and then use pip3 because it installed the wrong version and complained that the other versions in the requirements.txt didn't exist.
I am pretty sure I was in an activated venv, but there's a possibility I wasn't.

Phrozyn:
I'm going to close this until I can re-test.

PR #1390, opened 2019-07-22
updating alert plugin for py3.6

Phrozyn:


pwnbus:
I don’t think these changes are needed, dict.values returns an iterator so that in-line forloop should work without it

Phrozyn:
Closing if not needed.

PR #1391, opened 2019-07-22
adding list() where I think it may be needed in alerts

Phrozyn:


pwnbus:
These changes are not needed, we’re able to iterate just fine on dict.items like this

Phrozyn:
Closing if not needed.

PR #1400, opened 2019-07-29
Senctences wern't ending with dot in some places.

Siddharth1698:
Just observed that in README document some paragraphs wern't ending with dots.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ The Mozilla Enterprise Defense Platform (MozDef) seeks to automate the security 
 ## Goals:
 
 * Provide a platform for use by defenders to rapidly discover and respond to security incidents.
-* Automate interfaces to other systems like bunker, cymon, mig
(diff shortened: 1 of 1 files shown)


gene1wood:
Good catch on this inconsistency. I think in this case the best fix is to remove the period on the one that has one [as these aren't complete sentences nor completions of stems](https://www.businesswritingblog.com/business_writing/2012/01/punctuating-bullet-points-.html). I'll remove that period and check the other lists (as I've never really thought about this before, just sort of intuited what I should do (which turns out doesn't conform to what most writing style guides suggest).

Anyhow, again, thank you for catching this, really appreciate it.

Siddharth1698:
Sure... Always ready to help 😊

gene1wood:
@Siddharth1698 I've taken your suggested change, reversed it and applied it out to all our doc pages in #1401. Thanks again.

PR #1404, opened 2019-07-30
Resolves the tab-nav issue in themes.

Phrozyn:


meteor/imports/themes/classic/mozdef.css
```diff
@@ -287,19 +287,6 @@ caption, legend {
     color: var(--txt-secondary-color);
 }
 
-/* incident/investigation styles */
-.daterangepicker, .daterangepicker td {
```
Phrozyn: This was removed as we no longer have any daterangepicker elements in our html.

PR #1408, opened 2019-07-31
Etintel alert plugin

mpurzynski:


alerts/lib/etintellib.py
```diff
@@ -0,0 +1,52 @@
...
+        url = ''
+        self.rawdata = dict()
+        if netaddr.valid_ipv4(self.intel):
+            url = "https://" + self.options.apiurl + "/ips/" + self.intel + "/" + reqtype
+        else:
+            url = "https://" + self.options.apiurl + "/domains/" + self.intel + "/" + reqtype
+        if len(reqtype) and len(url) > 1:
+            try:
+                request = requests.get(url=url, headers=headers, timeout=2)
+                print(request)
```
pwnbus: Leftover from debugging?

alerts/lib/etintellib.py
```diff
@@ -0,0 +1,52 @@
...
+            except (
+                    requests.exceptions.ConnectionError, requests.exceptions.TooManyRedirects,
+                    requests.exceptions.Timeout) as e:
+                sys.stderr.write('Failed to fetch IQRisk data: {0}\n'.format(e))
```
pwnbus: We should use our logger class instead: `logger.error('Failed to fetch ...')`

alerts/lib/etintellib.py
```diff
@@ -0,0 +1,52 @@
...
+            if 'request' in locals() and hasattr(request, 'json'):
+                if request.status_code == 200:
+                    try:
+                        self.rawdata = request.json()
+                    except (ValueError) as e:
+                        sys.stderr.write('Failed to decode IQRisk data: {0}\n'.format(e))
```
pwnbus: Same comment here, use our logger class

alerts/lib/etintellib.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_reputation(self, type, objtype):
```
pwnbus: Can we rename `type` to be something else? It's a bit confusing because `type` is already "used" by python. IE: `type('abcd)`

alerts/plugins/etintel.py
```diff
@@ -0,0 +1,86 @@
...
+import sys
+from datetime import datetime, timedelta
+sys.path.append(os.path.join(os.path.dirname(__file__), "../lib"))
+from etintellib import etintel
+
+
+class message(object):
+    def __init__(self):
+        '''
+        takes an incoming alert
```
pwnbus: Incorrect comments

alerts/plugins/etintel.conf
```diff
@@ -0,0 +1,6 @@
+[options]
+keywords = sshd squid
+apiurl=api.emergingthreats.net/v1
+apikey=<SECRET>
+allowedipreptypes=reputation, urls, domains, events
```
pwnbus: Since this eventually gets split by `,`, we should remove the extra spaces here inbetween the commas to make it a bit easier and to avoid confusion for folks to customize.

alerts/plugins/etintel.py
```diff
@@ -0,0 +1,86 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import os
+from configlib import getConfig, OptionParser
+import sys
+from datetime import datetime, timedelta
+sys.path.append(os.path.join(os.path.dirname(__file__), "../lib"))
```
pwnbus: We're trying to get away from modifying the sys.path if possible, we can actually do something like:
`from lib.etintellib import etintel`

PR #1410, opened 2019-08-01
Fix port scan enrich

arcrose:
This PR fixes a problem in the port scan enrichment alert plugin that attempted to reference a field in alerts that does not exist.  Instead, we'll read the sourceip from the _first_ event (there is guaranteed to be at least one) recorded under the alert.


alerts/plugins/port_scan_enrichment.py
```diff
@@ -196,10 +196,10 @@ def enrich(alert, search_fn, search_window, max_connections):
         for hit in results.get('hits', [])
     ]
 
-    alert['details']['recentconnections'] = []
```
pwnbus: I think we still want to save the recentconnections array on alerts['details'], not the root.

PR #1414, opened 2019-08-07
Uptycs Alert Integration

claudijd:
This is just a stub to start with, which we can evolve and flesh out soonish.


cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # TODO: Fix this so hostName is properly set
+        mozmsg.log["hostname"] = alert["host"]
+        for item in alert:
+            details[item] = alert[item]
+
+        localdetails = normalize(details)
+        mozmsg.details = localdetails
+        mozmsg.summary = (
+            alert["eventtype"] + " " + alert["result"] + " for " + alert["username"]
```
claudijd: Need to clean this up

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+        return alerts_json['items']
+
+
+def normalize(details):
+    # Normalizes fields to conform to http://mozdef.readthedocs.io/en/latest/usage.html#mandatory-fields
+    # This is mainly used for common field names to put inside the details structure
+    # There might be faster ways to do this
+    normalized = {}
+
+    # for f in details:
```
claudijd: Need to review alert JSON values and understand what should be normalized.

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+        last_alert_ids = state["last_alert_ids"]
+    else:
+        last_alert_ids = []
+        filter.start_time = utc_now - datetime.timedelta(days=30)
+
+    filter.end_time = utc_now
+
+    # Query alerts that match the filter
+    alerts = client.alerts(filter)
+
```
claudijd: Need to re-add alert_ids de-dup logic.

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+                     (self.url, self.customer_id, path))
+        try:
+            response = requests.get(url=final_url,
+                                    headers=self.auth_headers(),
+                                    verify=self.verify_ssl)
+            if (response and response.status_code in
+                    [requests.codes.ok, requests.codes.bad]):
+                return response.json()
+        except requests.exceptions.RequestException as e:
+            print("ERROR: {}".format(e))
```
claudijd: I presume we'll want better error/exception handling here to reset and preserve our statefile.

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+    utc_now = datetime.datetime.utcnow()
+
+    # If an existing state file exists, set our start window to last run
+    if os.path.exists(options.statepath):
+        state = pickle.load(open(options.statepath, 'rb'))
+        filter.start_time = state["last_run"]
+        last_alert_ids = state["last_alert_ids"]
+    else:
+        last_alert_ids = []
+        filter.start_time = utc_now - datetime.timedelta(days=30)
```
claudijd: Need to move this initial bootstrap time into a config parameter

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,208 @@
...
+    #         normalized["username"] = normalized["user"]["name"]
+    #     if "key" in normalized["user"]:
+    #         normalized["userkey"] = normalized["user"]["key"]
+    #     del (normalized["user"])
+
+    # return normalized
+    return details
+
+
+def process_alerts(mozmsg, uptycs_alerts):
```
claudijd: Need to have maximum alerts to MozDef threshold to shut down the run and wait until the next execution time.

cron/uptycs_alertpull.py
```diff
@@ -0,0 +1,288 @@
...
+        dest="configfile",
+        default=defaultconfigfile,
+        help="configuration file to use",
+    )
+    (options, args) = parser.parse_args()
+    initConfig()
+    main()
+
+
+# Example Alert JSON
```
claudijd: @pwnbus here's a pretty print alert JSON example from Uptycs, please let me know if there are specific key fields you can immediately grok that we should normalize for and I can make the adjustments in code.

cron/uptycs_alertpull.py
claudijd: Saving this for a future iteration.

cron/uptycs_alertpull.py
pwnbus: I just ran it against our mapping from production and it appears that we don't have any conflicts, so we should be good to go!

PR #1424, opened 2019-08-20
Geomodel v0.1

arcrose:
This PR contains an implementation of GeoModel v0.1 modelled as an alert.


tests/alerts/geomodel/test_alert.py
```diff
@@ -0,0 +1,89 @@
+from datetime import datetime, timedelta
+import unittest
```
pwnbus: Since we use pytest to run our tests, we don't actually need to import unittest or inherit from unittest.TestCase, pytest will automatically detect and run it correctly.

docs/source/geomodel/specifications/v0_1.md
```diff
@@ -0,0 +1,290 @@
+# GeoModel Version 0.1 Specification
```
pwnbus: What if we renamed this to `docs/source/geomodel.md`? Then we could add a reference to it in the docs so that it would show up in https://mozdef.readthedocs.io/en/latest/, maybe under the `Usage` or `Advanced Settings`?

alerts/geomodel/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
```
pwnbus: I don't think we need to have this file filled with contents, it seems a bit weird to have a version and email defined for a part of mozdef like this.

tests/alerts/test_geomodel_alert.py
```diff
@@ -0,0 +1,138 @@
...
+    # state is established in Toronto for user tester2.
+    # Given the states described above, detecting activity from tester1 in
+    # Toronto is expected to trigger an alert since it is a new locality that
+    # tester1 could not have travelled to.  On the other hand, no alert should
+    # fire for tester2 since both localities are in Toronto.
+
+    default_event = {
+        '_source': {
+            'sourceipaddress': '1.2.3.4',
+            'utctimestamp': _now,
```
pwnbus: You can leave off this field, as it will get automatically added to any event as part of the test run.

alerts/geomodel_alert.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python
```
pwnbus: I think it seems a bit weird to have `_alert` be in the filename, but I think I'd agree with you renaming it away from geomodel (the same as your folder), what about something like `geomodel_location`, thoughts?

alerts/geomodel/alert.py
```diff
@@ -0,0 +1,135 @@
...
+        username: str,
+        sourceip: str,
+        origin: Origin,
+        summary: str = _DEFAULT_SUMMARY
+    ) -> 'Alert':
+        '''Produce a new `Alert` with default values filled.
+        '''
+
+        return Alert(
+            source='geomodel',
```
pwnbus: Many of these fields aren't used anywhere, (ex: source, type_, timestamp), and since we only use it in geomodel_alert as parameters to createAlertDict, what if we specified `tags` and `category` in the geomodel_alert, and used a dict for the other keys?

alerts/geomodel/config.py
```diff
@@ -0,0 +1,61 @@
...
+class Localities(NamedTuple):
+    '''Contains configuration for working with locality state.
+    '''
+
+    es_index: str
+    valid_duration_days: int
+    radius_kilometres: float
+
+
+class SearchWindow(NamedTuple):
```
pwnbus: Is a class needed here if we only have one field? What if https://github.com/mozilla/MozDef/pull/1424/files#diff-691e6b4ee17eb7e5df421c87f045908cR28 was `search_window: int`

alerts/geomodel_alert.json
```diff
@@ -0,0 +1,25 @@
+{
+  "elasticsearch_address": "",
+  "localities": {
+    "es_index": "localities",
+    "valid_duration_days": 1,
+    "radius_kilometres": 50
+  },
+  "events": [
+    {
+      "es_index": "events-*",
```
pwnbus: Instead of defining the es index in the config file here like this, we can reference this variable from the alert https://github.com/mozilla/MozDef/blob/master/alerts/lib/alerttask.py#L108

alerts/geomodel/locality.py
```diff
@@ -0,0 +1,255 @@
...
+
+    _source = event.get('_source', {})
+
+    source_ip = _source.get('sourceipaddress')
+    geo_data = _source.get('sourceipgeolocation')
+
+    if source_ip is None or geo_data is None:
+        return None
+
+    now = datetime.strftime(datetime.utcnow(), '%Y-%m-%dT%H:%M:%S.%f+00:00')
```
pwnbus: I think we should use the `toUTC` function https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/toUTC.py#L10  This line then becomes `now = toUTC(datetime.now()).isoformat()`

alerts/geomodel/locality.py
```diff
@@ -0,0 +1,255 @@
...
+    return Locality(source_ip, city, country, active_time, lat, lon, radius)
+
+
+def find(qes: QueryInterface, username: str, index: str) -> Optional[Entry]:
+    '''Retrieve the locality state for one user from ElasticSearch.
+    '''
+
+    search = SearchQuery()
+    search.add_must([
+        TermMatch('type_', 'locality'),
```
pwnbus: The convention is mozdef is to just use `type` here for things like this.

alerts/geomodel/locality.py
```diff
@@ -0,0 +1,255 @@
...
+    return Update(state, did_update)
+
+
+def remove_outdated(state: State, days_valid: int) -> Update:
+    '''Update a state by removing localities that are outdated, determined
+    by checking if the last activity within a given locality was at least
+    some number of days ago.
+    '''
+
+    now = datetime.utcnow().replace(tzinfo=pytz.UTC)
```
pwnbus: The timezone info is all contained within toUTC, so it's preferred to call that method instead. This line would then become `now = toUTC(datetime.now())`

alerts/geomodel/locality.py
```diff
@@ -0,0 +1,255 @@
...
+        for loc in state.localities
+        if loc.lastaction >= last_valid_date
+    ]
+
+    return Update(
+        state=State(state.type_, state.username, new_localities),
+        did_update=len(new_localities) != len(state.localities))
+
+
+def _parse_datetime(datetime_str):
```
pwnbus: We shouldn't need to specify this logic, the event timestamps should be all in a guessable format, so we can simply call `toUTC` from https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/toUTC.py#L10

alerts/geomodel/config.py
```diff
@@ -0,0 +1,61 @@
...
+    users: List[str]
+    cidrs: List[str]
+
+
+class Alerts(NamedTuple):
+    '''Contains configuration required to generate and store alerts in
+    ElasticSearch.
+    '''
+
+    es_index: str
```
pwnbus: We shouldn't be needing to know about the index for alerts, since none of the geomodel code is actually writing or reading from the alerts index.

alerts/geomodel_alert.json
```diff
@@ -0,0 +1,25 @@
...
+      "search_window": {
+        "minutes": 5
+      },
+      "lucene_query": "tags:auth0",
+      "username_path": "details.username"
+    }
+  ],
+  "alerts": {
+    "es_index": "alerts-*",
+    "whitelist": {
```
pwnbus: I think it's a bit weird to have the whitelist be under the alerts key, what if we made `whitelist` be a key under root?

tests/alerts/geomodel/test_locality.py
```diff
@@ -0,0 +1,270 @@
...
+    '''
+
+    def closure(q: SearchQuery, esi: str) -> Optional[locality.Entry]:
+        return results
+
+    return closure
+
+
+def _now() -> datetime:
+    return datetime.utcnow().replace(tzinfo=pytz.UTC)
```
pwnbus: We should use `toUTC` https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/toUTC.py#L10 instead

PR #1433, opened 2019-08-28
Fix alert schedules with modifiedat date as string

pwnbus:


pwnbus:
Closing in favor of https://github.com/mozilla/MozDef/pull/1437

Issue #1435, opened 2019-08-30
Alerts stop running if you disable one via web ui


pwnbus:
Now that our alerts schedule is pulled from restapi (mongodb), there's a nasty bug that will prevent all alerts from running, if one alert is disabled via the web ui.

PR #1437, opened 2019-09-04
Fix mongodb -> restapi -> celery task scheduler

pwnbus:
The previous iteration had a few bad bugs, so we had to revert the entire PR. This is a second pass at it, as I believe the bugs identified are now all fixed.


pwnbus:
Reapplies https://github.com/mozilla/MozDef/pull/1436 and some fixes on top of it.

pwnbus:
Fixes https://github.com/mozilla/MozDef/issues/1435

PR #1454, opened 2019-09-13
Stackdriver

mpurzynski:


mpurzynski:
RFC for now - looking for feedback

Issue #1189, opened 2019-03-28
Elasticsearch Docker Container configuration


Phrozyn:
We should look at disabling swap on the elasticsearch container.

- https://docs.docker.com/config/containers/resource_constraints/#prevent-a-container-from-using-swap

- add: bootstrap.memory_lock: false to the configuration file elasticsearch.yml.

Phrozyn:
Jira Task: EIS-753

Phrozyn:
This shouldn't affect cloudy mozdef as that is run using managed es in aws.
This is only really relevant to a docker dev instance if it is intended to use any significant amount of resources.

PR #1455, opened 2019-09-17
Disable memory lock in elasticsearch docker config

pwnbus:


pwnbus:
https://github.com/mozilla/MozDef/issues/1189

pwnbus:
Apparently according to docker-compose v3 (we use 3.7), there isn't a way to specifically disable swap? Regardless, I think it's fine to leave that ability out for now.

PR #1457, opened 2019-09-18
Fixup alert plugins

pwnbus:


alerts/plugins/port_scan_enrichment.json
```diff
@@ -1,5 +1,4 @@
 {
-  "elasticSearchAddress": "http://127.0.0.1:9200",
```
tristanweir: Why is this no longer configurable?

pwnbus: We inherit the elasticsearch address from the original alerts configuration (https://github.com/mozilla/MozDef/blob/master/alerts/lib/config.py#L40). I don't see there being a case where we would want to point to a different ES cluster for this specific plugin.

PR #1460, opened 2019-09-20
Process Windows Event Logs

mpurzynski:
Process Windows Event Logs - both local and forwarded if you so fancy.
Tested in a real-world Active Directory domain with several application servers and a dedicated Windows Event Collector via nxlog -> Fluentd (TCP input) -> SQS.
Will work with other transportation mechanisms.


mq/plugins/wec.py
```diff
@@ -0,0 +1,80 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# Copyright (c) 2017 Mozilla Corporation
+
+import os
+import jmespath
+import yaml
+from mozdef_util.utilities.toUTC import toUTC
+
+
+class message(object):
```
pwnbus: Can you add a few tests for this plugin?

PR #1464, opened 2019-09-24
Remove duplicate version of files in cloudymozdef build

pwnbus:


pwnbus:
@andrewkrug Since during setup, we copy the files into the build directory, I don't think we want to have them version controlled (as they are in the repo in other files). Thoughts?

andrewkrug:
Does the make target still work?

pwnbus:
@andrewkrug we should be good because we cp the files in https://github.com/mozilla/MozDef/blob/master/cloudy_mozdef/lambda_layer/Makefile#L30-L31

PR #1473, opened 2019-09-30
Add LDAP bruteforce alert

claudijd:
Makes use of soon to be deployed user attribute to offer per user bruteforce alerting in MozDef.  This should stay as a WIP until we have verifiable data in prod with the exact formats to spec out.

### DIFF
--- a/alerts/ldap_bruteforce.conf
+++ b/alerts/ldap_bruteforce.conf
@@ -1,0 +1,3 @@
+[options]
+threshold_count = 10
+search_depth_min = 120
--- a/alerts/ldap_bruteforce.conf
+++ b/alerts/ldap_bruteforce.conf
@@ -1,3 +1,4 @@
 [options]
 threshold_count = 1
 search_depth_min = 60
+host_exclusions = foo.example.com,bar.example.com 
--- a/alerts/ldap_bruteforce.conf
+++ b/alerts/ldap_bruteforce.conf
@@ -1,3 +1,3 @@
 [options]
-threshold_count = 10
-search_depth_min = 120
+threshold_count = 1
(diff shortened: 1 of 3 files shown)


alerts/ldap_bruteforce.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class AlertLdapBruteforce(AlertTask):
+    def main(self):
+        self.parse_config('ldap_bruteforce.conf', ['threshold_count', 'search_depth_min'])
+        search_query = SearchQuery(minutes=int(self.config.search_depth_min))
+        search_query.add_must([
+            TermMatch('category', 'ldap'),
+            TermMatch('details.response.error', 'LDAP_INVALID_CREDENTIALS'),
+            TermMatch('details.authenticated', 'False')
```
claudijd: Need to add a term match that requires details.user to be non-empty.  @pwnbus what's my best match object for that type of behavior?

pwnbus: You can do something like:
```
search_query.add_must_not(TermMatch('details.user', ''))
```

claudijd: Thanks, added!

alerts/ldap_bruteforce.conf
```diff
@@ -0,0 +1,3 @@
+[options]
+threshold_count = 1
```
claudijd: We need to set this to a higher value in prod, but we can use our data to support where exactly we should set this threshold per our specific environment.

claudijd:
@pwnbus this is ready, just not the config will need to be tweaked before deploying to prod.

PR #1477, opened 2019-10-01
Add reference to rsyslog and omhttp module

darakian:
Quick little PR adding a reference to the rsyslog module. I can also add some example config if you guys like, but let me know where I should stick that.


pwnbus:
I think what you've got is definitely sufficient, if you do want to add a sample config (IMO the documentation is good enough, and we don't need a sample config, but entirely up to you), it might make sense to put in https://github.com/mozilla/MozDef/tree/master/config.

darakian:
I'll leave it out then. If you ever start up a page on example configs I'll add it then :)

PR #1478, opened 2019-10-02
Add host exclusions to ldap password spray alert

claudijd:


alerts/ldap_password_spray.conf
```diff
@@ -1,3 +1,4 @@
 [options]
 threshold_count = 1
-search_depth_min = 60
\ No newline at end of file
+search_depth_min = 60
+host_exclusions = foo.example.com,bar.example.com
```
claudijd: This host_exclusion list will need to be swapped for LDAP stage hosts, which we can get from @jdow

alerts/ldap_password_spray.py
```diff
@@ -13,8 +13,12 @@
 
 class AlertLdapPasswordSpray(AlertTask):
     def main(self):
-        self.parse_config('ldap_password_spray.conf', ['threshold_count', 'search_depth_min'])
+        self.parse_config('ldap_password_spray.conf', ['threshold_count', 'search_depth_min', 'host_exclusions'])
         search_query = SearchQuery(minutes=int(self.config.search_depth_min))
+
+        for host_exclusion in self.config.host_exclusions.split(","):
```
pwnbus: Add an example test case that verifies this exclusion is working as expected?

claudijd: Added

PR #1479, opened 2019-10-02
adding mqPlugin 'zoom_fixup.py' to remove uneccessary fields

Phrozyn:
Removes topic

### DIFF
--- a/mq/plugins/zoom_fixup.py
+++ b/mq/plugins/zoom_fixup.py
@@ -1,0 +1,26 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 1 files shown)


mq/plugins/zoom_fixup.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class message(object):
+    def __init__(self):
+        '''register our criteria for being passed a message
+           as a list of lower case strings or values to match with an event's dictionary of keys or values
+           set the priority if you have a preference for order of plugins to run. 0 goes first, 100 is assumed/default if not sent
+        '''
+        # get auditd data
+        self.registration = ['zoom', 'mozdef-event-framework-zoom-dev']
```
pwnbus: Is `mozdef-event-framework-zoom-dev` leftover testing?

Phrozyn: yes, and so is the auditd comment, I'll fix that shortly.

PR #1480, opened 2019-10-03
Add whitelist to proxy drop ip alert

claudijd:
Adds the capability to white-list certain destinations to prevent alerting when they happen.  In the 169 example, it's a demonstration of a misconfiguration, but not  asecurity issue.  This can be produced in a reporting means for sysadmin tasks, but is not a security event because this IP isn't routable via the proxy.

### DIFF
--- a/alerts/proxy_drop_ip.py
+++ b/alerts/proxy_drop_ip.py
@@ -13,6 +13,8 @@ import netaddr
 
 class AlertProxyDropIP(AlertTask):
     def main(self):
+        self.parse_config("proxy_drop_ip.conf", ["ip_whitelist"])
+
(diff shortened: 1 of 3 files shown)


alerts/proxy_drop_ip.py
```diff
@@ -28,6 +30,9 @@ def main(self):
 
         search_query.add_must([QueryStringMatch("details.host: {}".format(ip_regex))])
 
+        for ip in self.config.ip_whitelist.split(","):
```
pwnbus: Can we add an example test that verifes we won't alert if the whitelist matches?

claudijd: Added a test for this.

PR #1481, opened 2019-10-03
Stackdriver final

mpurzynski:


mq/plugins/stackdriver.py
```diff
@@ -0,0 +1,68 @@
...
+class message(object):
+    def __init__(self):
+        """
+            Plugin used to fix object type discretions with cloudtrail messages
+        """
+        self.registration = ["pubsub"]
+        self.priority = 5
+
+        try:
+            self.mozdefhostname = "{0}".format(node())
```
pwnbus: This variable isn't used anymore it looks like.

mq/plugins/stackdriver.py
```diff
@@ -0,0 +1,68 @@
...
+            return (message, metadata)
+
+        event = message["details"]
+
+        if "logName" not in event:
+            return (message, metadata)
+        else:
+            # XXX: implement filtering of audit types that we want to see (yaml)
+            newmessage = dict()
+            logtype = urllib.parse.unquote(event["logName"]).split("/")[-1].strip()
```
pwnbus: Is there any chance that `event["logName"]` isn't this format? Should we verify that before we try and access the last element in the array? `[-1]`?

mq/plugins/stackdriver_audit.py
```diff
@@ -0,0 +1,102 @@
...
+class message(object):
+    def __init__(self):
+        """
+            Plugin used to fix object type discretions with cloudtrail messages
+        """
+        self.registration = ["stackdriver"]
+        self.priority = 15
+
+        try:
+            self.mozdefhostname = "{0}".format(node())
```
pwnbus: Similar comment as before, can we delete this block since the variable doesn't appear to be used anywhere?

mq/plugins/stackdriver_gceactivity.py
```diff
@@ -0,0 +1,72 @@
...
+class message(object):
+    def __init__(self):
+        """
+            Plugin used to fix object type discretions with cloudtrail messages
+        """
+        self.registration = ["stackdriver"]
+        self.priority = 16
+
+        try:
+            self.mozdefhostname = "{0}".format(node())
```
pwnbus: Similar comment about unused variable.

mq/plugins/stackdriver_syslog.py
```diff
@@ -0,0 +1,53 @@
...
+class message(object):
+    def __init__(self):
+        """
+            Plugin used to fix object type discretions with cloudtrail messages
+        """
+        self.registration = ["stackdriver"]
+        self.priority = 15
+
+        try:
+            self.mozdefhostname = "{0}".format(node())
```
pwnbus: Similar comment about unsed variable.

mq/plugins/stackdriver_syslog.py
```diff
@@ -0,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+# Copyright (c) 2017 Mozilla Corporation
+
+from mozdef_util.utilities.toUTC import toUTC
+from platform import node
+
+
+class message(object):
```
pwnbus: Can we add a few tests for each of the plugins?

mq/esworker_pubsub.py
```diff
@@ -0,0 +1,188 @@
...
+from mozdef_util.utilities.logger import logger, initLogger
+from mozdef_util.elasticsearch_client import (
+    ElasticsearchClient,
+    ElasticsearchBadServer,
+    ElasticsearchInvalidIndex,
+    ElasticsearchException,
+)
+from google.cloud import pubsub
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../"))
```
pwnbus: We're trying to get away from modifying sys.path, so this can be replaced with `from lib.plugins import sendEventToPlugins, registerPlugins`

mq/plugins/stackdriver.py
mpurzynski: I added a check for the field's presence, but I do not expect the format to change here. It is part of official event definition and Google seems to keep those intact. In other words, if this changes than it most likely means we have to write our plugins again, because they changed  data structures for everything they have.

mq/plugins/stackdriver_gceactivity.py
```diff
@@ -0,0 +1,63 @@
...
+        newmessage["details"]["gceactivity"] = newmessage["details"]["jsonPayload"]
+        del (newmessage["details"]["jsonPayload"])
+        newmessage["details"]["service"] = "compute.googleapis.com"
+
+        if message["category"] in self.eventtypes:
+            for key in self.yap[newmessage["category"]]:
+                mappedvalue = jmespath.search(self.yap[newmessage["category"]][key], newmessage)
+                # JMESPath likes to silently return a None object
+                if mappedvalue is not None:
+                    newmessage["details"][key] = mappedvalue
```
Phrozyn: These events do not appear to have a summary, please craft a summary for them.

mq/plugins/stackdriver_audit.py
```diff
@@ -0,0 +1,66 @@
...
+
+        todel = set()
+        if message["category"] in self.eventtypes:
+            for key in self.yap[newmessage["category"]]:
+                mappedvalue = jmespath.search(self.yap[newmessage["category"]][key], newmessage)
+                todel.add(self.yap[newmessage["category"]][key])
+                # JMESPath likes to silently return a None object
+                if mappedvalue is not None:
+                    newmessage["details"][key] = mappedvalue
+
```
Phrozyn: These events do not appear to have a summary, please craft a summary for them.

mq/plugins/stackdriver.py
```diff
@@ -0,0 +1,63 @@
...
+                        newmessage["source"] = "stackdriver"
+                        newmessage["tags"] = message["tags"] + ["stackdriver"]
+
+            newmessage["receivedtimestamp"] = toUTC(message["receivedtimestamp"]).isoformat()
+            newmessage["timestamp"] = toUTC(event["timestamp"]).isoformat()
+            newmessage["utctimestamp"] = toUTC(event["timestamp"]).isoformat()
+            newmessage["mozdefhostname"] = message["mozdefhostname"]
+            newmessage["customendpoint"] = ""
+            newmessage["details"] = event
+
```
Phrozyn: If you don't craft a summary here, please craft one in a fixup plugin as you did for the syslog plugin for stackdriver syslog events.

mq/plugins/stackdriver_gceactivity.py
mpurzynski: There's no value in having this field

mq/plugins/stackdriver.py
mpurzynski: The summary field cannot be crafted here by design

mq/plugins/stackdriver_audit.py
mpurzynski: There is no value in having a summary field

mq/esworker_pubsub.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, esConnection, options):
+        self.esConnection = esConnection
+        self.pluginList = registerPlugins()
+        self.options = options
+        self.scopes = ["https://www.googleapis.com/auth/cloud-platform", "https://www.googleapis.com/auth/pubsub "]
+        self.credentials_file = options.credentials_file
+
+    def run(self):
+        # XXX: fetch from the config file
+        subscriber = pubsub.SubscriberClient.from_service_account_file(self.options.credentials_file)
```
pwnbus: Should we have an example credentials file added to mozdef source, so we know what the format of that file is?

mq/esworker_pubsub.py
```diff
@@ -0,0 +1,185 @@
...
+    options.credentials_file = getConfig("credentials_file", "", options.configfile)
+
+    options.mqprotocol = getConfig("mqprotocol", "pubsub", options.configfile)
+
+    # plugin options
+    # secs to pass before checking for new/updated plugins
+    # seems to cause memory leaks..
+    # regular updates are disabled for now,
+    # though we set the frequency anyway.
+    options.plugincheckfrequency = getConfig("plugincheckfrequency", 120, options.configfile)
```
pwnbus: This option is no longer used, so we can delete it.

mq/esworker_pubsub.py
mpurzynski: We do most of the time, I would pass on it here. The reason being - the whole file is downloaded from the GCP console and used as it is.

mpurzynski: We could have a placeholder telling people to download the file with credentials.

mq/esworker_pubsub.py
mpurzynski: Removed

PR #1489, opened 2019-10-15
Enrich geomodel alert summary

arcrose:
Include the distance between hops as well as the time between when activity in each location was observed.


tests/alerts/test_geomodel_location.py
```diff
@@ -158,6 +111,7 @@ class TestAlertGeoModel(GeoModelTest):
                         'country': 'US',
                         'latitude': 37.773972,
                         'longitude': -122.431297,
+                        'observed': '2017-01-01T05:44:00+00:00',
```
pwnbus: Now that we're comparing timestamps when comparing the `default_alert` against what we get from the geomodel alert, we'll want to use `toUTC` to put the time into whatever timezone we're using (which I believe is why the tests are currently failing).

PR #1493, opened 2019-10-17
moves ldap details.tls to details.tls_encrypted as boolean

Phrozyn:
I figured it'd be better to not have to disparate fields called the same thing, moving this to details.tls_encrypted is descriptive and should not interfere with any other details.tls field.

### DIFF
--- a/mq/plugins/ldap_fixup.py
+++ b/mq/plugins/ldap_fixup.py
@@ -1,0 +1,31 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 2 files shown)


mq/plugins/ldap_fixup.py
```diff
@@ -0,0 +1,31 @@
...
+        # check for category like 'ldap'
+        if key_exists('category', message):
+            if message.get['category'] == "ldap":
+                if key_exists('details.tls', message):
+                    message['details.tls_encrypted'] = message['details.tls']
+                    del message['details']['tls']
+
+        if 'source' not in message:
+            message['source'] = "ldap"
+            return (message, metadata)
```
pwnbus: This return statement isn't necessary since we'll hit the one directly under this.

Phrozyn: removed! ty!

PR #1494, opened 2019-10-17
Rename ldap alerts

claudijd:
This simply renames the LDAP bruteforcing alerts per a conversation I had with brandon where we were both getting confused about which one was which.  This should make the alerts in source and when they fire more clear what exactly is happening.


Phrozyn:
Can we change the category to "bruteforce" as we do with other bruteforce alerts?
Thsi will allow the source IP to be blocked when the threshold is hit.

pwnbus:
@Phrozyn Done in https://github.com/mozilla/MozDef/pull/1494/commits/cd411f0a4e6998eb0c64996c1941d8e652bd2434

PR #1497, opened 2019-10-18
moves unknown success values to successful key, avoiding conflicts pr…

Phrozyn:
…eserving summary


Phrozyn:
Going to fix this a different way.

PR #1512, opened 2019-11-04
Adding summary to github events.

Phrozyn:


tests/mq/plugins/test_github_webhooks.py
```diff
@@ -1859,6 +1859,7 @@ def test_organization(self):
         self.verify_metadata(metadata)
         self.verify_meta(message, result)
         self.verify_actor(message, result)
+        self.verify_repo(message, result)
         self.verify_org(message, result)
         assert result['source'] == 'organization'
         assert result['details']['action'] == message['body']['action']
```
pwnbus: Should we add some checks to the tests to make sure the summary is set as expected?

Phrozyn: It will populate as expected as long as those verify tests pass, as the summary is built from the values in those fields (details.action, details. details.username, and details.source, etc).

Phrozyn: new commit to make this happen!

PR #1516, opened 2019-11-11
A new version of the guardduty plugin and a dedicated worker

mpurzynski:


mq/esworker_guardduty.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.save_event(event, metadata)
+
+    def on_message(self, message_raw):
+        if "Message" in message_raw:
+            message = json.loads(message_raw["Message"])
+            if "details" in message:
+                if "finding" in message["details"]:
+                    if "action" in message["details"]["finding"]:
+                        if "actionType" in message["details"]["finding"]["action"]:
```
pwnbus: We can use the `key_exists` function from https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/key_exists.py to clean up these if statements, it'll change it to:

`if key_exists('details.finding.action.actionType', message):`

mq/esworker_guardduty.py
```diff
@@ -0,0 +1,170 @@
...
+
+import sys
+import os
+import socket
+from configlib import getConfig, OptionParser
+from datetime import datetime
+from mozdef_util.utilities.toUTC import toUTC
+from mozdef_util.utilities.logger import logger, initLogger
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../"))
```
pwnbus: We're trying to get away from modifying the sys.path directly, so you shouldn't need to do this statement anymore (similarly to the other mq workers)

tests/mq/plugins/test_guardduty.py
```diff
@@ -0,0 +1,536 @@
...
...etworkConnectionAction': {'connectionDirection': 'INBOUND', 'remoteIpDetails': {'ipAddressV4': '198.51.100.0', 'organization': {'asn': '-1', 'asnOrg': 'GeneratedFindingASNOrg', 'isp': 'GeneratedFindingISP', 'org': 'GeneratedFindingORG'}, 'country': {'countryName': 'GeneratedFindingCountryName'}, 'city': {'cityName': 'GeneratedFindingCityName'}, 'geoLocation': {'lat': 0, 'lon': 0}}, 'remotePortDetails': {'port': 39677, 'portName': 'Unknown'}, 'localPortDetails': {'port': 80, 'portName': 'HTTP'}, 'protocol': 'TCP', 'blocked': False}}, 'resourceRole': 'TARGET', 'additionalInfo': {'sample': True}, 'evidence': {'threatIntelligenceDetails': [{'threatNames':['GeneratedFindingThreatName'], 'threatListName':'GeneratedFindingThreatListName'}]}, 'eventFirstSeen': '2019-08-08T20:09:13.816Z', 'eventLastSeen': '2019-10-25T00:24:41.104Z', 'archived': False, 'count': 21}, 'category': 'UnauthorizedAccess:EC2/TorIPCaller', 'tags': ['NETWORK_CONNECTION']}, 'hostname': 'i-99999999', 'summary': 'IP address 198.51.100.0 on the Tor Anonymizing Proxy network is communicating with EC2 instance i-99999999.', 'utctimestamp': '2019-10-25T00:26:17.227000+00:00', 'timestamp': '2019-10-25T00:26:17.227000+00:00'}
```
pwnbus: Can we break this out into multiple lines? Makes it much more consumable.

mq/plugins/guardDuty.py
```diff
@@ -3,87 +3,140 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 # Copyright (c) 2017 Mozilla Corporation
 
-from mozdef_util.utilities.key_exists import key_exists
+import os
```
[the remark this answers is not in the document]
mpurzynski: I wasn't using it in this module, I do use it in places where there's multiple levels of dictionaries to traverse, while making sure each intermediate level exists

mq/esworker_guardduty.py
```diff
@@ -0,0 +1,166 @@
...
+            event["tags"] = [self.options.taskexchange]
+
+        event["severity"] = "INFO"
+        event["source"] = "guardduty"
+        event["details"] = {}
+
+        event["details"] = message["details"]
+        if "hostname" in message:
+            event["hostname"] = message["hostname"]
+        if "summary" in message:
```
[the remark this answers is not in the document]
mpurzynski: Just trying to separate the transport layer from the actual data standardization. The new model looks like this:
- the gd worker extracts every single event from whatever AWS sends us
- for each event, plugin is called, to do necessary data transformation

PR #1522, opened 2019-11-20
wip on resolving missing usernames

Phrozyn:
This is a WIP as I work out the issues causing username to be missing from specific events


Phrozyn:
Will open a new one when it's ready

PR #1524, opened 2019-11-21
adding bottom margin to make pagination visible before footer

Phrozyn:
This should be tested on QA first

### DIFF
--- a/meteor/imports/themes/classic/mozdef.css
+++ b/meteor/imports/themes/classic/mozdef.css
@@ -57,6 +57,10 @@ body {
     color: black;
 }
 
+.container {
+    margin-bottom: 3%;
+}
+
 footer {
     text-align: center;
     position: fixed;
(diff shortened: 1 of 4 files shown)


meteor/imports/themes/classic/mozdef.css
```diff
@@ -57,6 +57,10 @@ body {
     color: black;
 }
 
+.container {
```
pwnbus: I think instead of adding style to the container class that is used by other areas of the web ui, we instead probably want to create a specific id. Currently this will cause headers to be pushed down the page:
Before:
Screen Shot 2019-11-21 at 10 46 34 AM

After:
Screen Shot 2019-11-21 at 10 46 25 AM

pwnbus: Decided to go ahead with this change and then we can focus on fixing up specifically the menu bar.

PR #1528, opened 2019-12-04
Alert action invoke

pwnbus:


alerts/lib/alerttask.py
```diff
@@ -546,3 +538,10 @@ def parse_json_alert_config(self, config_file):
                 logger.error("FAILED to open the configuration file\n")
 
         return json_obj
+
+    def generate_full_doc(self, alert_body, alert_es):
+        full_alert = {}
```
arcrose: nit:  We could write

```python
return {
    '_id': alert_es['_id'],
    '_index': alert_es['index'],
    '_source': alert_body
}
```

pwnbus: Agreed, that looks cleaner, added in https://github.com/mozilla/MozDef/pull/1528/commits/b180218dd782c53252118da2474e5a5fc18b46e2

bot/irc/mozdefbot.py
```diff
@@ -303,6 +303,8 @@ def on_message(self, body, message):
                     "alertworker exception: unknown body type received %r" % body)
                 return
 
+            bodyDict = full_body['_source']
```
arcrose: Nit: `bodyDict` (here) is called `body_dict` in `bot/slack/mozdefbot.py`.

Phrozyn: seconded, probably doesn't change the functionality, but consistency is nice.

pwnbus: Fixed in https://github.com/mozilla/MozDef/pull/1528/commits/85fd60c3929d62d4a5faa43a4d7841313b1aba38

PR #1531, opened 2019-12-12
Handle the case when a single API call name is sent as a json dict in…

mpurzynski:
…stead of a single-element list

### DIFF
--- a/mq/plugins/guardduty_mapping.yml
+++ b/mq/plugins/guardduty_mapping.yml
@@ -1048,18 +1048,15 @@
     awsaccountid: details.accountId
     awsregion: details.region
     resourcetype: details.resource.resourceType
-    accesskeyid: details.resource.accessKeyDetails.accessKeyId
-    principalid: details.resource.accessKeyDetails.principalId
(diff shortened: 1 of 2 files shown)


mq/plugins/guardduty_mapping.yml
```diff
@@ -1048,111 +1048,95 @@
...
   Recon:IAMUser/UserPermissions:
     findingid: details.id
     arn: details.arn
     awsaccountid: details.accountId
     awsregion: details.region
     resourcetype: details.resource.resourceType
-    accesskeyid: details.resource.accessKeyDetails.accessKeyId
-    principalid: details.resource.accessKeyDetails.principalId
-    usertype: details.resource.accessKeyDetails.userType
-    username: details.resource.accessKeyDetails.userName
```
tristanweir: Don't we need userName and userType in a UserPermissions event?

Phrozyn: @mpurzynski Do these specific types of events not have userName and userType mappings?

Phrozyn:
These mappings are not final but @mpurzynski has stated these will be covered in a future PR.

PR #1536, opened 2019-12-26
initial commit of Auth0 bruteforce user

claudijd:
Had a discussion with lucius about this, figured I'd just propose it while we were thinking about it.  I think it could be enriched further if we wanted to append the geo-data associated with the failure to help IR folks (or even computers) respond.

### DIFF
--- a/alerts/auth0_bruteforce_user.conf
+++ b/alerts/auth0_bruteforce_user.conf
@@ -1,0 +1,4 @@
+[options]
+threshold_count = 1
+search_depth_min = 60
+host_exclusions = foo.example.com,bar.example.com 
--- a/alerts/auth0_bruteforce_user.py
+++ b/alerts/auth0_bruteforce_user.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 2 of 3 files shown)


claudijd:
Ways this could be made cooler:

1.) Appending Geo Data to the alert summary ("Auth0 Bruteforce Attack in Progress against user (jsmith@example.com) from Berlin, Germany (1.2.3.4)")

2.) Appending Geo Data to the alert summary with history ("Auth0 Bruteforce Attack in Progress against user (jsmith@example.com) from Berlin, Germany (1.2.3.4) when user usually logs in from San Francisco, CA, US")

3.) Appending Geo Data to the alert summary with history and an action ("Auth0 Bruteforce Attack in Progress against user (jsmith@example.com) from Berlin, Germany (1.2.3.4) when user usually logs in from San Francisco, CA, US.  I've sent the user an email and a slack notification to let them know about the behavior and that the source IP has been blocked.")

Phrozyn:
So we have an LDAP bruteforce alert, I hesitate to say we also need an Auth0 one.
Instead I would think we'd want to enhance the current brute force alert we have to encompass the chain of authentication that takes place so we can have a single alert that includes the chain of events from all integrated sources.

Having another alert will result in more of the same we've been battling for a while, alert fatigue.

Phrozyn:
We are closing this PR as it has not had any activity in some time. Thank you for submitting, and if there is an opportunity in the future where this can be used, we can refer back to it.

PR #1546, opened 2020-01-15
Alert plugin enrich portscan

arcrose:
Not the best branch name ever, but the plugin itself is a little better.

I opted to create a new plugin for this functionality.  The existing enrichment plugins are poorly named considering the names don't say what they enrich such alerts _with_, and my goal with this one was to correct for that.  In the future, we may want to get DHCP-related information for other alerts, and this plugin would be a good place to add that in to.


arcrose:
This PR has been tested in QA and can be shown to be making the alert modifications we expect, as in the attached screenshot.
Screenshot from 2020-03-25 13-49-12

PR #1547, opened 2020-01-15
Fix the port_scan_enrichment alert plugin

arcrose:
I noticed while I was working on the other plugin I opened a PR for that this one was overriding `alert.details` with just the field it adds. This PR corrects that.


alerts/plugins/port_scan_enrichment.py
```diff
@@ -196,9 +196,7 @@ def enrich(alert, search_fn, search_window, max_connections):
         for hit in results.get('hits', [])
     ]
 
-    alert['details'] = {
-        'recentconnections': []
-    }
+    alert['details']['recentconnections'] = []
```
pwnbus: I think we also want to add a check to see if alert['details'] exists beforehand, and if it doesn't, we set it.

PR #1549, opened 2020-01-16
Allow meteor dev container with refresh on files

pwnbus:
This runs a dev meteor setup in a docker container (with `make run-dev-meteor`), and since we're mounting the local meteor directory, we can modify the files locally and the meteor container will autorefresh with the new file changes.

### DIFF
--- a/docker/compose/dev-meteor.yml
+++ b/docker/compose/dev-meteor.yml
@@ -8,7 +8,7 @@ services:
       args:
         METEOR_BUILD: 'NO'
     restart: always
-    command: bash -c 'node -i'
+    command: bash -c 'cd /opt/mozdef/envs/mozdef/meteor;meteor npm install --save @babel/runtime;meteor'
     depends_on:
       - mongodb
       - rest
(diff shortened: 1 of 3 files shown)


docker/compose/dev-meteor.yml
```diff
@@ -8,13 +8,11 @@ services:
...
     restart: always
-    command: bash -c 'node -i'
+    command: bash -c 'cd /opt/mozdef/envs/mozdef/meteor;meteor npm install --save @babel/runtime;meteor'
     depends_on:
       - mongodb
       - rest
     networks:
       - default
     volumes:
         - ../../meteor:/opt/mozdef/envs/mozdef/meteor
```
Phrozyn: if we are mounting a local volume, that path may be different on every workstation that runs this. Can we provide a comment at least to point out that if your local copy location is different to enter that path?

pwnbus: Assuming you checkout mozdef source and run the make target, this local path shouldn't change between user's dev environments, the second `/opt/mozdef/envs/mozdef/meteor` is the docker container's location, so that shouldn't be different between environments.

Phrozyn: Sounds good :)
 TY!

PR #1553, opened 2020-01-29
[Ready to Go] Fix geomodel noisiness

arcrose:
This PR includes additions to the geomodel_location alert that add support for the enrichment of the alert with information about external "factors," such as info. about ASNs that IP addresses in the alert belong to.  The first such factor implemented will raise the severity of the alert from INFO to WARNING in the case that any of the user in question's hops from one IP to another involve movement from one ASN to another.

Further work will need to be done to make this alert fully functional and ready for production:

1. A cron job to periodically download/update the MaxMind GeoIP ASN database needs to be implemented.
2. The question about how to manage the alert's management of a MaxMind Reader should be addressed.
3. Testing in QA should be performed to ensure that the alert still runs as expected and that the factor is correctly set up and invoked.

I'd like to call on @mpurzynski for support with the first. For the second, the question can be answered by @pwnbus.  As for the last item on the list, I believe the manual testing is something I, perhaps with the help of either @mpurzynski or @pwnbus, can do most of the work for.


arcrose:
I have tested that the modifications to this alert appear to work as expected in QA.

Below, you can see that when no movement from one ASN to another is detected, an alert's severity remains at the new default level INFO and there are no "factors" auxiliary data appended.

Screen Shot 2020-03-30 at 6 10 30 PM

In this second example, you can see that when movement from one ASN to another **is** detected, that the alert's severity is escalated to the WARNING level and that the auxiliary "factors" data pertaining to those ASNs is appended to the alert.

Screen Shot 2020-03-30 at 6 54 17 PM

I will address the issues in Travis shortly.

Phrozyn:
> * A cron job to periodically download/update the MaxMind GeoIP ASN database needs to be implemented.
> 
> * The question about how to manage the alert's management of a MaxMind Reader should be addressed.
> 
> * Testing in QA should be performed to ensure that the alert still runs as expected and that the factor is correctly set up and invoked.

@arcrose can you/have you created cards for this work to be done?
I don't want to merge this only to not have these requirements captured. These needs should also be voiced during sprint planning.

PR #1556, opened 2020-02-07
Remove confluence shell alert

pwnbus:


Phrozyn:
We will need to write a new alert for these since confluence is now hosted in AWS, but I expect the alert to search for different things, and can be created anew rather than using this once we determine the actual event details.

PR #1562, opened 2020-02-12
Extend geoip asn

mpurzynski:


cron/update_geolite_db.py
```diff
@@ -34,24 +36,29 @@ def fetch_db_data(db_download_location):
...
+        logger.debug("Saving db data to " + temp_path)
+        temp.write(db_data)
+        fsync(temp.fileno())
+        temp.flush()
+        logger.debug("Testing temp geolite db file")
+        geo_ip = GeoIP(temp_path)
+        # Do a generic lookup to verify we don't get any errors (malformed data)
+        geo_ip.lookup_ip('8.8.8.8')
+        logger.debug("Moving temp file to " + save_path)
+    temp.close()
```
arcrose: `temp` will be closed by the context manager that it exists inside of as a result of being created with `with`.  Calling this method a second time may cause an error.

cron/update_geolite_db.py
```diff
@@ -34,24 +36,29 @@ def fetch_db_data(db_download_location):
...
+    close(fd)
+    rename(temp_path, save_path)
 
 
 def main():
     logger.debug('Starting')
-    logger.debug(options)
-    db_data = fetch_db_data(options.db_download_location)
-    save_db_data(options.db_location, db_data)
+    for db_file in options.db_files.split(','):
```
arcrose: How do we want to respond to @pwnbus' feedback that we may prefer to store each of the values we're currently parsing out in their own variables?  Personally, I'm in favor of the idea of defining separate variables rather than implementing this parsing logic.

arcrose:
I have verified that this addition to the cron job, deployed with the changes in the associated mozdef-deploy PR, works as expected.

image

cron/update_geolite_db.py
```diff
@@ -60,9 +69,9 @@ def initConfig():
     options.sysloghostname = getConfig('sysloghostname', 'localhost', options.configfile)
     options.syslogport = getConfig('syslogport', 514, options.configfile)
 
-    options.db_download_location = getConfig('db_download_location', '', options.configfile)
-    options.db_location = getConfig('db_location', '', options.configfile)
-
+    options.db_store_location = getConfig('db_store_location', '', options.configfile)
```
Phrozyn, line 72: Here the options vars are renamed, but I don't see an associated config file in this PR to ensure these options are indeed renamed. 
Also to build this out in MozDef-deploy, is there an associated PR?

mpurzynski: Updated the example config file

PR #1568, opened 2020-02-19
Remove blue page visualizations

pwnbus:
This also removes the idea of an Attacker, in favor of using an alert action to do the blocking from https://github.com/mozilla/MozDef/pull/1567


jeffbryner:
😢

pwnbus:
@jeffbryner We're definitely pouring one out for the ogres :) 

We just felt that the kibana visualizations are the better, more supported route.

jeffbryner:
Understood, ogres aren't for everyone ;-] They'll live on in forks 🍴

Phrozyn:
Farewell dearest Ogres

PR #1576, opened 2020-03-11
Feature triage bot v02

arcrose:
This is a _huge_ PR with lots of redundant little bits of formatting changes and the like, so I'll provide a list here linking to parts of it that should be focused on during review:

* The entire [triage_bot.py](https://github.com/mozilla/MozDef/pull/1576/files#diff-2eee4bc268122543813936986382c732) alert action which manages duplicate chains that restrict when to send messages to the Slack bot Lambda function.
* [Tests](https://github.com/mozilla/MozDef/pull/1576/files#diff-b7f2b86f3cefe4bbe6906ebe9d6a1d12R303) for the triage bot alert action.
* The entire [triage_bot.py](https://github.com/mozilla/MozDef/pull/1576/files#diff-0639b5f3ae1eb6e99b5ec653ecc7811d) cron task which looks for duplicate chains whose primary alert have been responded to on Slack and replays the user's response to update the statuses of duplicate alerts.
* The entire [triage_bot.py](https://github.com/mozilla/MozDef/pull/1576/files#diff-a40507615cad5559b41757d1632cb10f) MQ plugin which updates the status of an alert upon receiving a response from a user on Slack.
* [Tests](https://github.com/mozilla/MozDef/pull/1576/files#diff-f76312d2013fa212e5e05c573a2867b1) for the triage bot MQ plugin.
* Additional [REST API endpoints](https://github.com/mozilla/MozDef/pull/1576/files#diff-c9dff09105e9694576c5a91b26bcd704R553) used to update alert statuses and manage duplicate chains.
* [Tests](https://github.com/mozilla/MozDef/pull/1576/files#diff-374e7714c27220b640238e8534da398e) for the additional REST API endpoints.
* The small new addition of the [insert_sqs.py](https://github.com/mozilla/MozDef/pull/1576/files#diff-3bc84006895ed7404f680fa283177e5b) script that sends messages into SQS.
* A small change I made to an existing [test_elasticsearch_client.py](https://github.com/mozilla/MozDef/pull/1576/files#diff-c2a231d4e0c306b8c6eec1c10496ce15) test that was failing around the time that a day changes before an index is created.


gene1wood:
I'm unsure how to approach reviewing a 3000 line 147 commit PR. Maybe a good place to start is to add some test to the description of this PR that talks about what this PR is all about (what feature it intends to add or bug it intends to fix)

arcrose:
Great point @gene1wood.  I've updated the opening comment for the PR with links to the most important things.

docker/compose/docker-compose-tests.yml
```diff
@@ -15,8 +15,8 @@ services:
       - base
     restart: always
     command: bin/elasticsearch
-    # ports:
-    #   - 9200:9200
+    ports:
```
pwnbus: I think we want this to remain commented out, so that you can run tests while also running an instance of mozdef in docker, which can be useful for doing dev. If you want to run ES + Rabbit exposed for testing purposes, you can run `make run-tests-resources-external`

cron/createFDQNBlockList.py
```diff
@@ -121,7 +121,7 @@ def initConfig():
 
     # mongo instance
     options.mongohost = getConfig('mongohost', 'localhost', options.configfile)
-    options.mongoport = getConfig('mongoport', 3001, options.configfile)
+    options.mongoport = getConfig('mongoport', 3002, options.configfile)
```
pwnbus: I don't know that these changes are necessary for this PR, so in the spirit of trying to minimize changes, I think we should keep the original default value.

alerts/actions/triage_bot.py
```diff
@@ -0,0 +1,746 @@
...
+CONFIG_FILE = os.path.join(os.path.dirname(__file__), "triage_bot.json")
+OAUTH_URL = "https://auth.mozilla.auth0.com/oauth/token"
+PERSON_API_BASE = "https://person-api.sso.mozilla.com"
+PERSON_API_AUDIENCE = "api.sso.mozilla.com"
+PERSON_API_SCOPE = (
+    "classification:public display:all display:public display:none search:all"
+)
+PERSON_API_GRANTS = "client_credentials"
+TOKEN_VALIDITY_WINDOW_MINUTES = 18 * 60
+SLACK_BOT_FUNCTION_NAME_PREFIX = "MozDefSlackTriageBotAPI-SlackTriageBotApiFunction"
```
pwnbus: These variables should get moved into the config file.

alerts/actions/triage_bot.py
```diff
@@ -0,0 +1,746 @@
...
+) -> types.Optional[AlertTriageRequest]:
+    """Attempt to determine the kind of alert contained in `alert` in
+    order to produce an `AlertTriageRequest` destined for the web server comp.
+    """
+
+    _source = alert.get("_source", {})
+    category = _source.get("category")
+    tags = _source.get("tags", [])
+
+    is_sensitive_host_access = "session" in tags and category == "session"
```
pwnbus: Is there a way that instead of hardcoding which alerts we handle in the code, can we abstract that away to the config file?

rest/index.py
```diff
@@ -22,62 +24,83 @@
...
+    """
+
+    OK = 200
+    BAD_REQUEST = 400
+    INTERNAL_ERROR = 500
 
 
 def enable_cors(fn):
-    ''' cors decorator for rest/ajax'''
+    """ cors decorator for rest/ajax"""
```
pwnbus: In the spirit of minimizing changes, can we revert the change from single quotes to double quotes as well as the line length fixes? There's a lot of additional changes in this file that aren't specifically related to the work done in the PR, so reviewing this is quite a pain currently.

alerts/actions/triage_bot.py
```diff
@@ -0,0 +1,746 @@
...
+
+        return None  # No duplicate chain found
+
+    ids = resp_data.get("identifiers", [])
+
+    if len(ids) == 0:
+        return None
+
+    try:
+        created = datetime.strptime(resp_data["created"], DUP_CHAIN_DATE_FMT)
```
pwnbus: Instead of needing to create DUP_CHAIN_DATE_FMT, we should use https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/toUTC.py#L10

rest/index.py
```diff
@@ -22,62 +24,83 @@
...
 
 options = None
-pluginList = list()   # tuple of module,registration dict,priority
+pluginList = list()  # tuple of module,registration dict,priority
+
+# The name of the MongoDB database that stores duplicate chains.
+DUP_CHAIN_DB = "duplicatechains"
+
+# The format string that UTC-based timestamps are encoded as.
+DUP_CHAIN_DATE_FMT = "%Y/%m/%d %H:%M:%S"
```
pwnbus: We should use https://github.com/mozilla/MozDef/blob/master/mozdef_util/mozdef_util/utilities/toUTC.py#L10 instead of needing to define the format like this.

scripts/ingest/insert_elasticsearch.py
```diff
@@ -31,32 +33,15 @@ def random_ip():
 # Fill in with events you want to write to elasticsearch
 # NEED TO MODIFY
 events = [
```
pwnbus: Can you re-add the example formatted event from before, so we know what format this event should be in when using it? The existing example event is also an event that will trigger an alert in mozdef docker environment, so having it hardcoded in the file in SCM is also useful.

tests/mozdef_util/test_elasticsearch_client.py
```diff
@@ -412,8 +412,8 @@ def test_get_indices(self):
         open_indices = self.es_client.get_open_indices()
         open_indices.sort()
         expected_indices = [self.alert_index_name, self.previous_event_index_name, self.event_index_name, 'test_index']
-        assert all_indices == expected_indices
-        assert open_indices == expected_indices
+        assert all([expected in all_indices for expected in expected_indices])
```
pwnbus: I don't think we need to make this change, do you have an example of this test failing? In a clean test environment, we shouldn't have any additional indices being created that we don't know about.

alerts/actions/triage_bot.py
arcrose: I believe we talked about this a while ago and concluded that there's no generic way of handling this with data alone.  Each alert that the bot is able to process has to follow some distinct steps to retrieve information about a user (duo alerts provide their email directly while others require a query to the PersonAPI for example). So, at some level, we always need a code addition to map some identifying information in an alert to a function that can process the information in that alert and collect additional information necessary to triage it.

rest/index.py
arcrose: I'm somewhat inclined to suggest we leave these changes in.  The bulk of these changes resulted from running the [black](https://github.com/psf/black) code formatter on this and several other files.  This tremendously helped to eliminate flake errors as well as to make code consistent.  I'd like to argue this is a good step towards making code style & formatting consistent, and I don't think it would be a good use of time to manually comb through the code to re-introduce non-conformant formatting.  The links I've provided in my opening comment should direct you to the portions of code that are critical to the requested review.

Do let me know if you disagree.

alerts/actions/triage_bot.py
arcrose: Would this change also necessitate moving from `datetime.utcnow()` to `datetime.now()` or is that fine as is?

tests/mozdef_util/test_elasticsearch_client.py
arcrose: I've pointed you to an example of this that I discovered on Slack last week.

rest/index.py
pwnbus: I disagree, I see the value with running the black code formatter on the codebase, but not as part of these specific changes, as it makes reviewing this PR significantly more difficult. These files were previously passing our flake8 tests, so there shouldn't have been previous errors (excluding the 5 flake8 tests we currently specifically exclude in https://github.com/mozilla/MozDef/blob/master/.flake8#L18-L22).

I think the work of :
  1. Reverting the file back to the state before this PR
  2. Manually copying/pasting the new code required for this PR
  3. Committing file
Would significantly help speed up the review process.

alerts/actions/triage_bot.py
pwnbus: It shouldn't be a problem, because while `datetime.now()` doesn't include the timezone, we automatically assume it's UTC in toUTC if a timezone isn't specified.

docker/compose/docker-compose-tests.yml
arcrose: Addressed in [recent commit](https://github.com/mozilla/MozDef/pull/1576/commits/bf3c595806d80188101aebcdec9b61a183d51c33).

cron/createFDQNBlockList.py
arcrose: Resolved in a [recent commit](https://github.com/mozilla/MozDef/pull/1576/commits/775a593f8c59f00026d4ff96e955d56b6a82a8bf).

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
+
+    details = {
+        "triage": {"user": req["user"], "response": req["response"]},
+        "identityConfidence": req["identityConfidence"],
+    }
+
+    modified_count = 0
+
+    modified_count += alerts.update_one(
+        {"esmetadata.id": req["alert"]}, {"$set": {"status": req["status"]}}
```
gene1wood, line 575: You may want to explicitly check for the presence of these fields to avoid a malformed JSON payload from causing an exception.

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
+            "error": msg,
+            "identifiers": [],
+            "created": toUTC(datetime.utcnow()).isoformat(),
+            "modified": toUTC(datetime.utcnow()).isoformat(),
+        })
+
+    query = {"alert": request.query.alert, "user": request.query.user}
+
+    if query["alert"] is None or query["user"] is None:
+        response.status = StatusCode.BAD_REQUEST
```
gene1wood, line 648: I think what you want to be testing for here [is an empty string, not `None`](https://bottlepy.org/docs/dev/api.html#bottle.FormsDict) because `FormsDict` returns empty strings on missing values.

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
+    response.content_type = "application/json"
+
+    initConfig()
+
+    mongo = MongoClient(options.mongohost, options.mongoport)
+    dupchains = mongo.meteor[DUP_CHAIN_DB]
+
+    try:
+        req = json.loads(request.body.read())
+        request.body.close()
```
gene1wood, line 719: You could use [bottle's native support for POSTed JSON](https://bottlepy.org/docs/dev/api.html#bottle.BaseRequest.json)

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
```
gene1wood, line 794: You could use [bottle's native support for POSTed JSON](https://bottlepy.org/docs/dev/api.html#bottle.BaseRequest.json)

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
+
+    mongo = MongoClient(options.mongohost, options.mongoport)
+    alerts = mongo.meteor["alerts"]
+
+    try:
+        req = json.loads(request.body.read())
+        request.body.close()
+    except ValueError:
+        response.status = StatusCode.BAD_REQUEST
+        response.body = json.dumps({"error": "Missing or invalid request body"})
```
gene1wood, line 542: In all of the cases in this file where `response.body` is set to a value that's returned from `json.dumps` I believe you can just use the [native bottle support for JSON by returning a dict](https://bottlepy.org/docs/dev/tutorial.html#generating-content).

```suggestion
        response.body = {"error": "Missing or invalid request body"}
```

I'll won't add this note to all the instances like this in the file but it should apply to all of them.

rest/index.py
```diff
@@ -468,6 +481,433 @@ def createIncident():
...
+
+    response.content_type = "application/json"
+
+    initConfig()
+
+    mongo = MongoClient(options.mongohost, options.mongoport)
+    dupchains = mongo.meteor[DUP_CHAIN_DB]
+
+    try:
+        req = json.loads(request.body.read())
```
gene1wood, line 882: You could use [bottle's native support for POSTed JSON](https://bottlepy.org/docs/dev/api.html#bottle.BaseRequest.json)

arcrose:
I appreciate the feedback!

I've addressed each of your requests in a handful of commits that I've just pushed up.

PR #1577, opened 2020-03-16
Fix test index set

arcrose:
This PR includes two changes:

1. An addition to our Makefile to clean docker before running tests and
2. A modification to a test in `test_elasticsearch_client.py` that will guarantee that, if execution of tests occurs across a change from one day to the next, that checks for indices will work as expected and include the new index created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ else
 endif
 
 .PHONY: test
-test: clean build-tests run-tests
+test: build-tests run-tests
 
 .PHONY: tests
 tests: build-tests run-tests  ## Run all tests (getting/building images as needed)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -40,7 +40,7 @@ else
 endif
 
 .PHONY: test
-test: build-tests run-tests
+test: clean build-tests run-tests
```
pwnbus: As our tests don't create any volumes, we shouldn't need to run the clean target as well.

Makefile
```diff
@@ -61,6 +61,7 @@ run-test: run-tests
 .PHONY: run-tests
 run-tests: run-tests-resources  ## Just run the tests (no build/get). Use `make TEST_CASE=tests/...` for specific tests only
 	docker run -it --rm mozdef/mozdef_tester bash -c "source /opt/mozdef/envs/python/bin/activate && flake8 --config .flake8 ./"
+	#docker run -it --rm --network=test-mozdef_default mozdef/mozdef_tester bash
```
pwnbus: Can we remove this commented out line?

PR #1598, opened 2020-04-06
Alert plugin possible usernames

arcrose:
This alert plugin registers on the `PromiscKernel` alert and enriches the alert with a `details.possible_usernames` field.  Each of the possible usernames listed are pulled from Syslog events indicating that an SSH authentication was accepted to the host that the `PromiscKernel` alert pertains to.

Given SSH authentication events such as the following

Screen Shot 2020-04-06 at 5 49 19 PM

the alert is enriched like so

Screen Shot 2020-04-06 at 5 48 33 PM


alerts/plugins/possible_usernames.json
```diff
@@ -0,0 +1,4 @@
+{
+  "search_window_hours": 24,
+  "indices_to_search": ["events-*"]
```
tristanweir, line 3: It might be more efficient to just search on `events` and `events-previous`, although that assumes a search window of 24 hours max.

arcrose: Given that this is a configuration file, these values are intended to be easy to change.  Want to leave these as the defaults if for no other reason than because it's consistent with other defaults?

alerts/plugins/possible_usernames.py
```diff
@@ -0,0 +1,104 @@
...
+
+        self._config = Config.load(CONFIG_FILE)
+
+        self._es_client = ElasticsearchClient(ES['servers'])
+
+    def onMessage(self, message):
+        summary_parts = message['summary'].split(' ')
+
+        # We expect the summary to be formatted like:
+        # "Promiscuous mode enabled on HOSTNAME [X]"
```
pwnbus, line 56: I think we should parse out the hostname via the events within the alert (which will most likely be one hostname inside many of the events), instead of parsing it out of the summary field. This way, we could make the plugin accessible to other alerts in the future, by just simply changing the registration

arcrose: I did consider this approach but my concern was that there's a potential, hypothetically, to retrieve a list of unique hostnames of length greater than one following this approach.  If that were the case, would it necessarily (in the case of all alerts) be safe to assume that either the hostname that appears most frequently is the one to choose?

pwnbus: Good question, yeah I think that's a good idea, if we do run into a situation with multiple different hostnames, use the most frequent one as what we search on.

arcrose: Got this working in 4c3d2fa

PR #1609, opened 2020-04-21
duo event parsing fixes

Phrozyn:
- add duo api host to hostname rather than mozdefhostname
- replace UTC logic with toUTC lib
- account for differences between api versions 1 and 2 in normalize function
- Flesh out summaries for administration events.

Did not see any events for telephony in any of the open indices we currently have, so those summaries are not fleshed out.

### DIFF
--- a/cron/duo_logpull.py
+++ b/cron/duo_logpull.py
@@ -5,30 +5,12 @@
 #
 
 import sys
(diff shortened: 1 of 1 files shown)


Phrozyn:
A future effort should be to create a test for this, I have created jira# EIS-2024 for that as a separate effort.

PR #1612, opened 2020-04-23
Remove sample alerts from demo

pwnbus:
Removed these alerts as they were created for demo purposes and is no longer valid.


pwnbus:
@Phrozyn Updated.

PR #1613, opened 2020-04-23
Remove auditd_commands alert

pwnbus:
Removed alert as this is no longer valid, and was discussed in our Alert Triage meetings.


pwnbus:
@Phrozyn Description updated.

PR #1614, opened 2020-04-24
Fix geomodel alert

arcrose:
The GeoModel alert was found to be stemming a runtime error in production as a result of `dict2List` producing in invalid output (a list containing complex data types that should have been decomposed).  This PR introduces two test cases for `dict2List`; one that constructs an obvious example of a manually-constructed data type with nested data structures and complex (tuple) keys as well as one test case taken directly from the output of a GeoModel alert.

In order to resolve the issue we observed, I have taken the approach of rewriting `dict2List` more or less from scratch.  The previous implementation was missing some special cases due to the way that it replicated processing, which is resolved by taking better advantage of recursion.  Here I use the `yield from` syntax [introduced in Python 3.3](https://docs.python.org/3/whatsnew/3.3.html#pep-380).


arcrose:
I've done some fairly extensive testing in QA to verify this change works as expected.  I performed the following procedure:

1. Modify `/opt/mozdef/envs/python/lib/python3.6/site-packages/mozdef_util/utilities/dict2List.py` so that it contains
   the new implementation of `dict2List`.
2. Import `dict2List` in an interpreter and verified that `list(dict2List({('complex', 'key'): 32}))`
   output the expected value of `['complex', 'key', 32]` so we know the right implementation is being used.
3. Triggered a GeoModel alert while the `geomodel_ipintel_enrichment` plugin is enabled.
4. Verified that the alert fires on well-formed events.
5. Used `insert_loginput.py` to insert to the log ingest to MozDef.
6. Verified that no errors were found in `eventtask.log`.
7. Verified that the events expected showed up in Kibana.

PR #1646, opened 2020-06-23
Alert when the Session Invalidation application is used to terminate a user's sessions

arcrose:
When the session invalidation application logs a message indicating that a user A has terminated sessions for user B, this alert fires with information about those users and their sessions.  In this case, there are duplicates because I ran my test a couple times with the same target user.

Screen Shot 2020-06-23 at 4 33 49 PM

### DIFF
--- a/alerts/session_invalidation.py
+++ b/alerts/session_invalidation.py
@@ -62,6 +62,8 @@ class AlertSessionInvalidation(AlertTask):
         alert = self.createAlertDict(summary, category, tags, events)
 
         alert['details'].update({
+            'actor': actor,
+            'username': actor,
             'terminations': terminations,
         })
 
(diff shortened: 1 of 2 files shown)


tristanweir:
Do these events fire on trigger or success? If it only happens when a responder triggers an invalidation, do we want to adjust any of the language to reflect that it might not have been successful (e.g. `{actor} attempted to invalidate session for {users}`?

arcrose:
The [code that emits the log](https://github.com/mozilla/session-invalidation/blob/master/lambda.py#L505) upon which this alert fires does so after session termination takes place, regardless of whether those terminations succeeded or failed.  So it is possible that some failed and others succeeded or that none will have succeeded at all.  I'm definitely open to adding some more specific cases to this but I feel like as far as the summary goes, this generic one is sufficient.  We had talked before about wanting to know when someone is using the app at all as well as when they terminate one or more sessions, and I figured this case captured both of those circumstances pretty well.  What do you think?

tristanweir:
>  We had talked before about wanting to know when someone is using the app at all as well as when they terminate one or more sessions, and I figured this case captured both of those circumstances pretty well.

Ok, sounds good to me

PR #1649, opened 2020-06-30
Cloudtrail parse errorfix

Phrozyn:
eis-2202 removes dots from beginning or end of a key name


Phrozyn:
Closing, will submit a new PR

PR #1650, opened 2020-06-30
remove dots from the start or end of a key

Phrozyn:
eis-2202 removes dots from beginning and/or end of a key name

### DIFF
--- a/mq/plugins/remove_dots.py
+++ b/mq/plugins/remove_dots.py
@@ -20,7 +20,7 @@ class message(object):
             if isinstance(message, dict):
                 message_keys = list(message.keys())
                 for key in message_keys:
-                    if key[0] == '.' or key[-1] == '.':
+                    if len(key) > 0 and (key[0] == '.' or key[-1] == '.'):
                         new_key = key.replace(".", "")
                         if new_key != key:
                             message[new_key] = message.pop(key)
--- a/mq/plugins/remove_dots.py
+++ b/mq/plugins/remove_dots.py
@@ -3,8 +3,6 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 # Copyright (c) 2014 Mozilla Corporation
 
-import json
-
 
 class message(object):
 
@@ -20,16 +18,21 @@ class message(object):
         self.priority = 5
 
     def onMessage(self, message, metadata):
-        def remove_dots(obj):
-            for key in obj.keys():
-                if key[0] == '.' or key[-1] == '.':
-                    new_key = key.replace(".","")
-                    if new_key != key:
-                        obj[new_key] = obj[key]
-                        del obj[key]
-            return obj
+        def renameKeysToRemoveDots(message):
+            if isinstance(message, dict):
+                message_keys = list(message.keys())
+                for key in message_keys:
+                    print(key)
+                    if key[0] == '.' or key[-1] == '.':
+                        new_key = key.replace(".","")
+                        if new_key != key:
+                            message[new_key] = message.pop(key)
+                    if isinstance(message.get(key), dict) or isinstance(message.get(key), list):
+                        message[key] = renameKeysToRemoveDots(message[key])
+            elif isinstance(message, list):
+                for item in message:
+                    item = renameKeysToRemoveDots(item)
+            return message
 
-        new_json = json.loads(json.dumps(message), object_hook=remove_dots)
-        message = new_json
-
+        message = renameKeysToRemoveDots(message)
         return (message, metadata)
--- a/mq/plugins/remove_dots.py
+++ b/mq/plugins/remove_dots.py
@@ -8,10 +8,8 @@ class message(object):
 
     def __init__(self):
         '''
-        takes an incoming message
-        and checks for dots at the
-        start or end of the key and
-        removes them
+        takes an incoming message and checks for dots at the
+        start or end of the key and removes them
         '''
 
         self.registration = ['cloudtrail']
@@ -22,12 +20,11 @@ class message(object):
             if isinstance(message, dict):
                 message_keys = list(message.keys())
                 for key in message_keys:
-                    print(key)
                     if key[0] == '.' or key[-1] == '.':
-                        new_key = key.replace(".","")
+                        new_key = key.replace(".", "")
                         if new_key != key:
                             message[new_key] = message.pop(key)
-                    if isinstance(message.get(key), dict) or isinstance(message.get(key), list):
+                    if isinstance(message.get(key), (dict, list)):
                         message[key] = renameKeysToRemoveDots(message[key])
             elif isinstance(message, list):
                 for item in message:
--- a/mq/plugins/remove_dots.py
+++ b/mq/plugins/remove_dots.py
@@ -1,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+import json
+
+
+class message(object):
+
+    def __init__(self):
+        '''
+        takes an incoming message
+        and sets the keys to lowercase
+        '''
+
+        self.registration = ['cloudtrail']
+        self.priority = 5
+
+    def onMessage(self, message, metadata):
+        def remove_dots(obj):
+            for key in obj.keys():
+                if key[0] == '.' or key[-1] == '.':
+                    new_key = key.replace(".","")
+                    if new_key != key:
+                        obj[new_key] = obj[key]
+                        del obj[key]
+            return obj
+
+        new_json = json.loads(json.dumps(message), object_hook=remove_dots)
+        message = new_json
+
+        return (message, metadata)
--- a/mq/plugins/remove_dots.py
+++ b/mq/plugins/remove_dots.py
@@ -11,7 +11,9 @@ class message(object):
     def __init__(self):
         '''
         takes an incoming message
-        and sets the keys to lowercase
+        and checks for dots at the
+        start or end of the key and
+        removes them
         '''
 
         self.registration = ['cloudtrail']
--- a/tests/mq/plugins/test_remove_dots.py
+++ b/tests/mq/plugins/test_remove_dots.py
@@ -65,3 +65,39 @@ class TestRemoveDotsPlugin():
         }
         assert retmessage == expected_message
         assert retmeta == {}
+
+    def test_list_dot_removal(self):
+        msg = {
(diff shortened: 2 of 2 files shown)


mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,35 @@
...
+        def remove_dots(obj):
+            for key in obj.keys():
+                if key[0] == '.' or key[-1] == '.':
+                    new_key = key.replace(".","")
+                    if new_key != key:
+                        obj[new_key] = obj[key]
+                        del obj[key]
+            return obj
+
+        new_json = json.loads(json.dumps(message), object_hook=remove_dots)
```
pwnbus, line 32: I'm a bit worried that doing both a dumps and a loads operation is a bit unnecessary and might cause performance problems. Instead of using json, would it be better to recursively loop through all the keys, similar to https://github.com/mozilla/MozDef/blob/master/mq/plugins/lower_keys.py . Thoughts?

Phrozyn: acknowledged, pinging you on slack

Phrozyn: I've  rewritten this code, I think I overcomplicated it when I originally attempted to make it similar to lower_keys, which is why I ended up with the dumps and loads.
This wasn't nearly as complicated this time around. Thank you!

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+
+        self.registration = ['cloudtrail']
+        self.priority = 5
+
+    def onMessage(self, message, metadata):
+        def renameKeysToRemoveDots(message):
+            if isinstance(message, dict):
+                message_keys = list(message.keys())
+                for key in message_keys:
+                    print(key)
```
pwnbus, line 25: Leftover debugging?

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+                    if key[0] == '.' or key[-1] == '.':
```
arcrose, line 26: I'd suggest using `str.startswith('.')` and `str.endswith('.')` here rather than string indexing, as it works safely with empty strings.

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+                        new_key = key.replace(".","")
+                        if new_key != key:
+                            message[new_key] = message.pop(key)
+                    if isinstance(message.get(key), dict) or isinstance(message.get(key), list):
+                        message[key] = renameKeysToRemoveDots(message[key])
+            elif isinstance(message, list):
+                for item in message:
+                    item = renameKeysToRemoveDots(item)
```
arcrose, line 34: There should be test cases for handling of all types of collections we deal with here.  This code, for example, would do nothing to modify message.  Example:

```py
>>> x = [".test", "test."]
>>>
>>> for item in x:
...   item = item.replace('.', '')
...
>>>
>>> x
['.test', 'test.']
```

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+        removes them
+        '''
+
+        self.registration = ['cloudtrail']
+        self.priority = 5
+
+    def onMessage(self, message, metadata):
+        def renameKeysToRemoveDots(message):
+            if isinstance(message, dict):
+                message_keys = list(message.keys())
```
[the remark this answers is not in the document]
arcrose: Also, you might want to consider writing

```py
for (key, value) in message.items():
```

to save from having to write `message[key]` all the time to get `value`.

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+# Copyright (c) 2014 Mozilla Corporation
+
+
+class message(object):
+
+    def __init__(self):
+        '''
+        takes an incoming message
```
arcrose, line 11: nit: these lines could be longer

mq/plugins/remove_dots.py
Phrozyn: lol yes

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def onMessage(self, message, metadata):
+        def renameKeysToRemoveDots(message):
+            if isinstance(message, dict):
+                message_keys = list(message.keys())
+                for key in message_keys:
+                    print(key)
+                    if key[0] == '.' or key[-1] == '.':
+                        new_key = key.replace(".","")
+                        if new_key != key:
```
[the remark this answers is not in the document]
Phrozyn: Is there something I'm missing in your feedback? There is a test case, and this section tests for beginning or end character and passes it to the next line only if it matches a dot, and then it only replaces a dot, it's discriminate.

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+                            message[new_key] = message.pop(key)
+                    if isinstance(message.get(key), dict) or isinstance(message.get(key), list):
```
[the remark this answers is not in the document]
Phrozyn: Good call

mq/plugins/remove_dots.py
Phrozyn: Thank you for this, I was unaware of this ability!

Phrozyn: Doing it this way actually adds to the processing time this takes.

here's a small example:
In this test I'm working with the following initial output: 
output = {'key0': {'key1': 'value2', '.key2.': ['this', 'list'] , 'key3': {'.ds_store': 'value4', 'key5 with a .': 'value5'}}}

runtime using indices:
{'ds_store': 'value4', 'key5 with a ': 'value5'}
{'key1': 'value2', 'key3': None, 'key2': ['this', 'list']}
{'key0': None}
---- 0.00017595291137695312 seconds ----

runtime using startswith and endswith builtins:
{'ds_store': 'value4', 'key5 with a ': 'value5'}
{'key1': 'value2', 'key3': None, 'key2': ['this', 'list']}
{'key0': None}
---- 7.295608520507812e-05 seconds ----

I'll keep using the indices as otherwise we might start seeing this bottleneck.

mq/plugins/remove_dots.py
```diff
@@ -0,0 +1,38 @@
...
+        self.priority = 5
+
+    def onMessage(self, message, metadata):
+        def renameKeysToRemoveDots(message):
+            if isinstance(message, dict):
+                message_keys = list(message.keys())
+                for key in message_keys:
+                    print(key)
+                    if key[0] == '.' or key[-1] == '.':
+                        new_key = key.replace(".","")
```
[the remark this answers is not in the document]
Phrozyn: noted

mq/plugins/remove_dots.py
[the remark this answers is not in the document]
Phrozyn: I can't duplicate what you've suggested here as being a problem in my testing.

mq/plugins/remove_dots.py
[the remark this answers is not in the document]
Phrozyn: I'm not working with the values, I'm only checking key contents to see if they are an instance of a dict or list. Using message.get(key) is also a safe way of obtaining that information that doesn't throw key errors.

mq/plugins/remove_dots.py
Phrozyn: noted, this was the original format of all templates :)

mq/plugins/remove_dots.py
Phrozyn: I've retested several times, and when doing a bazillion comparisons the indice is 2.5 times faster meaning the cycles we save in the interest of event processing is not worth the readability that startswith/endswith gives us.

mq/plugins/remove_dots.py
Phrozyn: I'll add a list to the test cases :)  it does work, contrary to what your snippet shows.

Phrozyn: So I dug into this a little more, and while it's true, list items are not keys, so they aren't really subject to this thing, it did get me to reviewing the code a bit more. I did add a test case for a key that starts a list and has dots at  the start and end of the key. This works as expected.

However, for dictionaries within a list, it will not work for those, and that is quite alright, because any list is going to be indexed as an object, which means a nested dict within a list will be a part of that object, and will not be parsed into a key value pair.

tests/mq/plugins/test_remove_dots.py
```diff
@@ -0,0 +1,67 @@
...
+                'responseelements': {
+                    'fileuploadurls': {
+                        'ds_store': 'astringvalue',
+                        'js/app.js': 'astringvalue'
+                    }
+                }
+            }
+        }
+        assert retmessage == expected_message
+        assert retmeta == {}
```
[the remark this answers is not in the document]
Phrozyn: Done

mq/plugins/remove_dots.py
arcrose: My concern is less with the readability and actually more the fault-tolerance. As I mentioned, if the length of the string happens to be 0 then using indices directly will result in a runtime exception.  Another way around this would be to check the length of the string ahead of time, writing something like

```py
if len(s) > 0 and (s[0] == '.' or s[-1] == '.'):
```

but that might have a similar or worse impact on the performance.

With that said I also don't think that squeezing fractions of performance is more important that writing more correct or readable code.  Not at this scale.

More to the point, though, it seems like using the startswith and endsiwth builtin is faster.  I guess this was an oversight because of the exponent?  0.00017 is 1.7e-04 and

```
>>> 7.295608520507812e-05 < 0.00017595291137695312
True
```

Phrozyn: When I made lower_keys I was asked to do testing on the time it takes before we could push that one to prod. That’s where this idea of performant came from.

here's more testing numbers for you so you can see exactly, I unfortunately had use two different methods for the numbers above:
phroztbyte:MozDef asmith$ python -m timeit -s 'text="foo"' 'text.startswith("a")'
2000000 loops, best of 5: 123 nsec per loop
phroztbyte:MozDef asmith$ python -m timeit -s 'text="foo"' 'text[0]=="a"'
5000000 loops, best of 5: 47.8 nsec per loop


We run every cloudtrail event through this, and at times we can have hundreds of thousands going through the pipe within minutes, this is why it can be critical

Phrozyn: I'll add the check for the key being null, but honestly since we are creating a list of keys that are present, there should never be an empty key, and if there was an empty key, it'd throw a parsing error way before this code is hit.

Issue #1607, opened 2020-04-17
Standardize "raw" field name and format


gene1wood:
Currently various plugins and cron fetchers use different field names when capturing a copy of the raw message that's being parsed. They also use different methods to serialize the data. Some use `str()`, others `json.dumps()` or the built in `repr` method.

# `rawdetails`
[broFixup.py](https://github.com/mozilla/MozDef/blob/6ede727ee2c34ccf61ef196c7645bca74f4952a4/mq/plugins/broFixup.py#L83)
[suricataFixup.py](https://github.com/mozilla/MozDef/blob/6ede727ee2c34ccf61ef196c7645bca74f4952a4/mq/plugins/suricataFixup.py#L50)

# `raw_value`
[cloudtrail.py](https://github.com/mozilla/MozDef/blob/6ede727ee2c34ccf61ef196c7645bca74f4952a4/mq/plugins/cloudtrail.py#L78)

# `raw`
[auth02mozdef.py](https://github.com/mozilla/MozDef/blob/6ede727ee2c34ccf61ef196c7645bca74f4952a4/cron/auth02mozdef.py#L357)

Would it make sense to standardize on one of these names so these raw messages all use the same field and store data in the same format?

Would a good format for that to be stored in (since this is stored in elastic search) be JSON? If so we should remember to set these field types from `t` to `?` so ES knows they contain JSON

Phrozyn:
I absolutely agree we should standardize, but the disparity comes from lack of formalization of a standard and different developers writing the different modules. @pwnbus I'd love to hear your thoughts!

gene1wood:
@pwnbus What do you think about this? What's a good next step on this (e.g. open a Jira ticket? assign this to someone?)

Phrozyn:
This is something we will tackle with our upcoming normalization efforts.
I'll add this to my requirements and plan. 
:) Thank you, Gene, for reminding me of this with your update!

gene1wood:
If/when #1653 is merged then there are only two raw field names, `raw_value` and `rawdetails`

PR #1653, opened 2020-07-06
Change Auth0 raw field from str repr to json

gene1wood:
This will both change the Auth0 raw field to be called raw_value
and will change the representation from a Python repr to a JSON
encoded string. This will allow ElasticSearch to index this value
and make it searchable.

This is similar to what was done in #1586 and support #1607

### DIFF
--- a/cron/auth02mozdef.py
+++ b/cron/auth02mozdef.py
@@ -9,6 +9,7 @@
 import hjson
 import sys
 import os
+import json
 import requests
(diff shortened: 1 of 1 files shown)


Phrozyn:
@gene1wood Travis isn't returning the completed results for the testing, when this happens I usually  have to close the PR and re-open, it's a weird bug with Travis.

PR #1656, opened 2020-07-08
Set an alert to acknowledged when the triage bot updates its status

arcrose:
When the triage bot updates an alert's status to "acknowledged" in response to a user indicating that an alert was expected via Slack, we previously only updated the new "status" field.  This PR introduces a small change that results in the triage bot MQ plugin that reads user responses adding the "acknowledged" and "acknowledgedby" fields to the alert in MongoDB so that it will show up as "ack'd" in the MozDef UI.

### DIFF
--- a/rest/index.py
+++ b/rest/index.py
@@ -579,17 +579,22 @@ def update_alert_status():
         "identityConfidence": req.get("identityConfidence"),
     }
 
-    modified_count = 0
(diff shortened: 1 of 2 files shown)


arcrose:
This is now running on our QA2 host and, as I've indicated in our private Slack channel, can be shown to be working as expected.

PR #1661, opened 2020-07-16
truncates htmlpart to something ES can handle and adds authparameters to modify_keys

Phrozyn:


mq/plugins/cloudtrail.py
```diff
@@ -117,6 +118,10 @@ def onMessage(self, message, metadata):
         if not message['source'] == 'cloudtrail':
             return (message, metadata)
 
+        if 'requestparameters' in message['details']:
```
arcrose, line 121: A simpler way of writing these two lines would be to do something like this:
```py
if message.get('details', {}).get('requestparameters', {}).get('htmlpart') is not None:
```

mq/plugins/cloudtrail.py
```diff
@@ -117,6 +118,10 @@ def onMessage(self, message, metadata):
...
+            if 'htmlpart' in message['details']['requestparameters']:
+                message['details']['requestparameters']['htmlpart'] = message['details']['requestparameters']['htmlpart'][0:4095]
```
arcrose, line 123: Could we define a constant that describes the meaning of the value 4095 here?

mq/plugins/cloudtrail.py
Phrozyn: This isn't consistent with how the rest of our code is currently written, but I'll change it.

mq/plugins/cloudtrail.py
Phrozyn: I'll add a comment to the code to help.

mq/plugins/cloudtrail.py
arcrose: I wouldn't consider this required; just a suggestion.

PR #1665, opened 2020-07-24
EIS-2365

Phrozyn:
Still a WIP

- Adds user email to administrative events that affect a user.
- Adds mfa_provider to details
- Adds user_last_ip
- Adds user_last_login
- Adds ldap_groups as a searchable object
- Adds msg as raw_value and a searchable object
- Adds success to successful API operations
- Makes summaries more informative
- removes json import, no longer needed
- Adds RP's to Summary

### DIFF
--- a/cron/auth02mozdef.py
+++ b/cron/auth02mozdef.py
@@ -168,7 +168,8 @@ def process_msg(mozmsg, msg):
         pass
 
     try:
-        details["username"] = msg.user_name
+        if msg.user_name:
+            details["username"] = msg.user_name
     except KeyError:
         pass
 
@@ -193,17 +194,20 @@ def process_msg(mozmsg, msg):
         pass
 
     try:
-        details["clientname"] = msg.client_name
+        if msg.client_name:
+            details["clientname"] = msg.client_name
     except KeyError:
         pass
 
     try:
-        details["connection"] = msg.connection
+        if msg.connection:
+            details["connection"] = msg.connection
     except KeyError:
         pass
 
     try:
-        details["clientid"] = msg.client_id
+        if msg.client_id:
+            details["clientid"] = msg.client_id
     except KeyError:
         pass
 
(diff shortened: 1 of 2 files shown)


cron/auth02mozdef.py
```diff
@@ -224,37 +245,50 @@ def process_msg(mozmsg, msg):
             # Update a rule, add a site, update a site, etc
             details["description"] = msg.description
     except KeyError:
-        details["description"] = ""
+        pass
```
tristanweir, line 248: Just so I understand the logic here, a KeyError would default to the loaded value becoming an empty string anyway? Or would it default to the value not existing at all?

If the second, have we verified that there are no alerts or other code that requires a details.description?

cron/auth02mozdef.py
```diff
@@ -224,37 +245,50 @@ def process_msg(mozmsg, msg):
...
-    else:
-        # default summary as action and description (if it exists)
-        mozmsg.summary = "{event} {desc}".format(event=details.eventname, desc=details.description)
+    # make summary be action/username (success login user@place.com)
+    # if no details.username field exists we don't add it.
+
+    # Build summary if neither email, description, nor username exists
+    if 'eventname' in details:
+        mozmsg.summary = "{event}".format(event=details.eventname)
+        if 'description' in details and details['description'] != "None":
```
tristanweir, line 257: Syntactically, are we looking for the string "None" here or are we looking for an empty value? I don't think we often populate a value with "None"

cron/auth02mozdef.py
Phrozyn: This was written this way due to the way the summary was originally built, description needed to exist as they did no checks to determine if it was populated. With the summary change, this is no longer a required field when empty.

cron/auth02mozdef.py
Phrozyn: We are looking for the string "None" as that is what Auth0 is sending us. A field with a value of the string "None"
I mentioned this in the slack channel, that this was one of the things that was tripping up summary building until I modified how we do it.

tristanweir: 👍

PR #1677, opened 2020-09-09
Fix #1676

### LINKED ISSUE #1676: Remove docker cron volume
What is the use of the **cron** volume (in docker/compose/docker-compose.yml) ? All the files in this volume are either in MozDef/cron or in MozDef/docker/compose/mozdef_cron/files, so it doesn't seem necessary to persist them as they are copied into the image when rebuilding the container.

The issue with this volume is that it prevents the modification of config files.
Example : if you modify _docker/compose/mozdef_cron/files/rotateIndexes.conf_ and rebuild the docker-compose cron service, it appears during the build that the new file is copied into the image, but when you run it with docker-compose and launch a bash in the container (`docker exec -it <cron_container> bash`) and print the _rotateIndexes.conf_ file (`cat opt/mozdef/envs/mozdef/cron/rotateIndexes.conf`), you'll see that it's the old config file.
That's because the mounting of the cron volume replaced the new file in the image, with the old version in the volume. (So you need to modify the config file in the volume for the change to take effect).

That makes it confusing as to why the changes in these config files do not take effect when running the container.
A solution could be to remove the cron volume.

matletix:


matletix:
Travis check failed because of #1678

pwnbus (approved):

PR #1694, opened 2020-12-14
Give endpoint methods unique names

matletix:
Give endpoint methods unique names so that they can be imported by other modules.


pwnbus (approved):

pwnbus:
Looks great, thanks for doing this! (And I apologize for the time it's taken to get a review and merged :\ )

Issue #363, opened 2016-07-13
Change plugin registration criteria comparison method to be more deterministic


gene1wood:
Currently the MozDef plugin system decides if an event matches the registered criteria of a given plugin by looking for any intersection between two lists. The first list is the `self.registration` list in the plugin ([example](https://github.com/mozilla/MozDef/blob/91889eca5bfa678984d62cf9b862908ad335c08b/alerts/plugins/pagerDutyTriggerEvent.py#L24)). The second list is a [collection of all the keys and all the values of all the nested dictionaries](https://github.com/mozilla/MozDef/blob/91889eca5bfa678984d62cf9b862908ad335c08b/alerts/alertWorker.py#L111) that makeup the event.

The rationale behind this method is so that a plugin could register a keyword and match that keyword against either
- a field name in an event
- a value in the tag field in an event

The downside to this method is that the plugin could match more criteria than intended. For example if the plugin author wants all events with a field named `http_x_forwarded_for` in the event, it's possible that an event with no `http_x_forwarded_for` field but with a `details` field with a value of `http_x_forwarded_for` would match.

A better approach may be to do two intentional checks. First check for any intersection between the registered plugin keyword list and the list of all dictionary keys in all nested dictionaries. Second check for any intersection between the registered plugin keyword list and the value of the tag field.

gene1wood:
Or we could enable plugin authors to be explicit in their registration declaration about their intent to match field names or tag values.

PR #1700, opened 2020-12-31
Add a stricter MQ plugin registration method

matletix:
The current method for registering a plugin is by passing a **list** of terms that should match keys or values of the processed events.
This additional registration method allows us to pass a **dict** that must be a sub-dict of the processed events for them to match.
That allows greater control on which events are processed by the plugins.

### DIFF
--- a/docs/source/plugins.rst
+++ b/docs/source/plugins.rst
@@ -1,7 +1,7 @@
 Plugins
 =======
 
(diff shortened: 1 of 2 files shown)


matletix:
Related to #363

matletix:
This PR addresses the MQ plugins. If it is of interest, it can be extended to the alerts plugins.

